PR #39, opened 2016-07-19
Comment and whitespaces handling in sql translation

diegosalvi:
Handles cases like `/* something */ CREATE TABLESPACE mytablespace` and similar issues


eolivelli:
what about creating an utility class since the rewrite query code is quite large now ?

diegosalvi:
I have moved _generic_ code to SQLUtils

PR #84, opened 2017-02-17
Add divide, multiply, abs, round

caiok:


caiok:
Pushed directly. PR no longer needed.

PR #176, opened 2018-02-16
reate new utils module

eolivelli:
The idea is to create a new module with utility and low level functions.
This will ease modularity and we will ease narrowing code coverage report on low level functions


eolivelli:
@diegosalvi  I would prefer to keep this change only as a directory move, leaving further changes to a separate commit, does this make sense to you ?

PR #177, opened 2018-02-24
Fix BLink lock excalation from read to write - condition not rechecked after upgrading lock

eolivelli:
This PR add a fix for BLink lock excalation.
There are additional tests and a NetBeans configuration file


eolivelli:
Okay, Closing

PR #178, opened 2018-03-01
Save resources by not writing tablespace uuid into WAL

eolivelli:


herddb-core/src/main/java/herddb/log/LogEntry.java
```diff
@@ -73,45 +86,45 @@ public int serialize(ExtendedDataOutputStream doo) throws IOException {
...
         doo.writeLong(transactionId); // 8
-        doo.writeUTF(tableSpace); // var
+        doo.writeUTF(""); // keep compatibility with v 0.2
         switch (type) {
             case LogEntryType.UPDATE:
                 doo.writeUTF(tableName);
                 doo.writeArray(key);
                 doo.writeArray(value);
-                return 18 + key.length + value.length + tableSpace.length() + tableName.length();
+                return 18 + key.length + value.length + EMPTY_STRING_LEN + WRITE_UTF_HEADER_LEN + tableName.length();
```
diegosalvi: Could it be done as following?

```
public LogEntry(long timestamp, short type, long transactionId, String tableName, byte[] key, byte[] value) {
    int startSize = doo.size();
    [...]
    return doo.size() - startSize;
    [...]
}
```

herddb-core/src/main/java/herddb/log/LogEntry.java
```diff
@@ -73,45 +86,45 @@ public int serialize(ExtendedDataOutputStream doo) throws IOException {
         doo.writeLong(timestamp); // 8
         doo.writeShort(type);  // 2
         doo.writeLong(transactionId); // 8
-        doo.writeUTF(tableSpace); // var
+        doo.writeUTF(""); // keep compatibility with v 0.2
```
diegosalvi: If an empty UTF string has already been evaluated in 38-51 why not writing directly its byte array without engoding it another time?

eolivelli:
@diegosalvi I have addressed your comments, please look

PR #181, opened 2018-03-02
Introduce wait in Planner if the default tablespace is not available

eolivelli:


eolivelli:
@diegosalvi  sorry the diff includes a reformat

Issue #182, opened 2018-03-05
Deadlock in BLink page unload


diegosalvi:
Many time we need to unload a blink page to load another one.

A deadlock occurs sometime during unload for pattern like:

- Thread A owns read load lock for page 1 and require write load lock to unload page 2
- Thread B owns read load lock for page 2 and require write load lock to unload page 1

PR #183, opened 2018-03-05
Fix BLink page unloading deadlock #182

diegosalvi:


herddb-utils/src/main/java/herddb/index/blink/BLink.java
```diff
@@ -2145,41 +2251,68 @@ boolean remove_link(X s, Node<X,Y> child) throws IOException {
...
             }
 
-            return new LockAndUnload<>(read, unload);
+            return new LockAndUnload<>(read, unload, pageId);
         }
 
+        boolean unload(boolean flush, boolean justTry) {
 
-        boolean unload(boolean flush) {
+            if (DEBUG)
```
eolivelli: nit: always use braches to enclose code blocks

PR #187, opened 2018-03-12
Use global lock to protect checkpoints

eolivelli:
We need to wait for all operation to end before starting a checkpoint


herddb-core/src/main/java/herddb/core/TableSpaceManager.java
```diff
@@ -117,6 +117,7 @@
 import herddb.storage.DataStorageManagerException;
 import herddb.storage.FullTableScanConsumer;
 import herddb.utils.Bytes;
+import java.util.concurrent.locks.StampedLock;
```
diegosalvi: StampedLock non viene poi utilizzato

Issue #198, opened 2018-05-30
BlockRangeIndex stack overflow


diegosalvi:
Running BlockRangeIndexBench I get a StackOverflow:

`java.lang.StackOverflowError
	at java.util.Objects.hashCode(Objects.java:98)
	at herddb.index.brin.BlockRangeIndex$BlockStartKey.hashCode(BlockRangeIndex.java:122)
	at java.util.HashMap.hash(HashMap.java:339)
	at java.util.HashMap.put(HashMap.java:612)
	at java.util.HashSet.add(HashSet.java:220)
	at herddb.index.brin.BlockRangeIndex$Block.lookUpRange(BlockRangeIndex.java:363)
	at herddb.index.brin.BlockRangeIndex$Block.lookUpRange(BlockRangeIndex.java:402)
	at herddb.index.brin.BlockRangeIndex$Block.lookUpRange(BlockRangeIndex.java:402)
        [...]
	at herddb.index.brin.BlockRangeIndex$Block.lookUpRange(BlockRangeIndex.java:402)
	at herddb.index.brin.BlockRangeIndex$Block.lookUpRange(BlockRangeIndex.java:402)
	at herddb.index.brin.BlockRangeIndex$Block.lookUpRange(BlockRangeIndex.java:402)
`

It could be better if lookUpRange don't invoke itself.

Possible improvements:
* do not invoke ricorsively, return a result list next node to continue by iteration
* do not chain streams each time, create a resulting list and create a stream from that (we could use an helper wrapper list to generate an unique stream or something similar)

PR #199, opened 2018-05-31
Fix BlockRangeIndex stack overflow, improve index seek times on delete and find (#198)

diegosalvi:
Fixed overflow, removed ricorsion from delete and lookup methods.
Removed mutable first key (now is immutable)

I think maxKey can be removed too but will require other work.

### DIFF
--- a/herddb-core/src/test/java/herddb/index/brin/BlockRangeIndexConcurrentTest.java
+++ b/herddb-core/src/test/java/herddb/index/brin/BlockRangeIndexConcurrentTest.java
@@ -93,16 +93,34 @@ public class BlockRangeIndexConcurrentTest {
     private void verifyIndex(BlockRangeIndex<Sized<Integer>, Sized<String>> index) {
         Integer lastmax = null;
         for (Block b : index.getBlocks().values()) {
-            System.out.println("check block " + lastmax + " -> " + ((Sized<Integer>) b.minKey).dummy + "," + ((Sized<Integer>) b.maxKey).dummy);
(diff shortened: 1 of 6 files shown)


herddb-core/src/test/java/herddb/index/brin/BlockRangeIndexConcurrentTest.java
```diff
@@ -93,16 +93,34 @@ private void dumpIndex(BlockRangeIndex<?, ?> index) {
     private void verifyIndex(BlockRangeIndex<Sized<Integer>, Sized<String>> index) {
         Integer lastmax = null;
         for (Block b : index.getBlocks().values()) {
-            System.out.println("check block " + lastmax + " -> " + ((Sized<Integer>) b.minKey).dummy + "," + ((Sized<Integer>) b.maxKey).dummy);
-            if (lastmax == null) {
-                lastmax = ((Sized<Integer>) b.maxKey).dummy;
+            if (b.key == BlockRangeIndex.BlockStartKey.HEAD_KEY.HEAD_KEY) {
+                System.out.println("check block " + lastmax + " -> -inf," + ((Sized<Integer>) b.maxKey).dummy);
```
eolivelli: maybe system.out is slowing down the test and introduces implicit barriers.
can we remove it ?

diegosalvi: No is on a "verify" test method ran at test end to check the index status. No concurrency nor updates is running at such time

PR #208, opened 2018-06-11
Perform early validation about indexed columns and clean up AbstractIndexManager interface

eolivelli:
Try to create indexed keys before accepting DML operations, in order to prevent data integrity problems.

Cleanup AbstractIndexManager and pass directly the Key of the index instead of the full record.

### DIFF
--- a/herddb-core/src/main/java/herddb/codec/RecordSerializer.java
+++ b/herddb-core/src/main/java/herddb/codec/RecordSerializer.java
@@ -521,8 +521,7 @@ public final class RecordSerializer {
         return toRecord(record, table);
     }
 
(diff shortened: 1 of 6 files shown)


herddb-core/src/main/java/herddb/codec/RecordSerializer.java
```diff
@@ -537,6 +536,7 @@ public static Bytes serializePrimaryKey(Map<String, Object> record, ColumnsList
             byte[] fieldValue = serialize(v, c.type);
             return new Bytes(fieldValue);
         } else {
+            ByteArrayOutputStream key = new ByteArrayOutputStream();
```
diegosalvi: nice catch

Issue #230, opened 2018-07-11
"Select *" does not work on Heap tables using JDBC


eolivelli:
The problem is that DataAccessorForFullRecord assumes implicitly that primary key columns are always the first columns on the table definition, this is not true in particular for HEAP tables (tables without PK)

Reproducer:
```
CREATE TABLE mytable (id string)
INSERT INTO mytable values('foo')
SELECT * FROM mytable
```

test case inside JDBC module:
```

@Test
    public void testSingleColumnHeapTable() throws Exception {
        try (Server server = new Server(new ServerConfiguration(folder.newFolder().toPath()))) {
            server.start();
            server.waitForStandaloneBoot();

            try (HDBClient client = new HDBClient(new ClientConfiguration(folder.newFolder().toPath()))) {
                client.setClientSideMetadataProvider(new StaticClientSideMetadataProvider(server));

                try (BasicHerdDBDataSource dataSource = new BasicHerdDBDataSource(client)) {

                    try (Connection con = dataSource.getConnection();
                            Statement statement = con.createStatement();) {
                        statement.execute("CREATE TABLE mytable (id string)");

                    }

                    try (Connection con = dataSource.getConnection();
                            PreparedStatement statement = con.prepareStatement("INSERT INTO mytable(id) values(?)");) {

                        for (int n = 0; n < 10; ++n) {
                            int i = 1;
                            statement.setString(i++, "mykey_" + n);
                            statement.addBatch();
                        }

                        int[] batches = statement.executeBatch();
                        Assert.assertEquals(10, batches.length);

                        for (int batch : batches) {
                            Assert.assertEquals(1, batch);
                        }
                    }

                    try (Connection con = dataSource.getConnection()) {

                        try (PreparedStatement statement = con.prepareStatement("SELECT * FROM mytable")) {
                            int count = 0;
                            try (ResultSet rs = statement.executeQuery()) {
                                while (rs.next()) {
                                    assertNotNull(rs.getString(1));
                                    ++count;
                                }
                            }
                            assertEquals(10, count);
                        }

                    }

                }
            }
        }
    }
```

PR #232, opened 2018-07-16
Issue-230 'Select *' does not work on Heap tables using JDBC

eolivelli:
this is the fix for #230.
Just put primary key columns as first columns in the Schema we are providing to Calcite.
This way DataAccessForFullRecord can leverage natural sort of data in memory


eolivelli:
@diegosalvi all tests passed on CI, this patch is ready to go for me

https://dev.majordodo.org/jenkins/job/HerdDB/1149/

PR #235, opened 2018-07-30
Add ability to invoke close on some embedded datasources on jdbc driver

diegosalvi:
Add some ability to close datasources  embedded into JDBC driver without completely unload the driver itself

### DIFF
--- a/herddb-jdbc/src/main/java/herddb/jdbc/Driver.java
+++ b/herddb-jdbc/src/main/java/herddb/jdbc/Driver.java
@@ -19,16 +19,20 @@
  */
 package herddb.jdbc;
 
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.DriverPropertyInfo;
 import java.sql.SQLException;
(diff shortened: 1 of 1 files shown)


herddb-jdbc/src/main/java/herddb/jdbc/Driver.java
```diff
@@ -120,4 +124,19 @@ public synchronized void close() {
         datasources.clear();
     }
 
+    /**
+     * Closes drive embedded datasources by their connection url
+     */
+    public synchronized void closeDatasources(String url) {
```
eolivelli: This is not cleaer to me.
The best way is to call "unregister" on the JDBC driver, this will call "close()" which will close all the datasources retained by the driver itself.

```
List<Driver> dd = Collections.list(DriverManager.getDrivers());
            dd.stream().forEach((d) -> {
                try {
                    System.out.println("[JDBC] unregister driver " + d.getClass());
                    DriverManager.deregisterDriver(d);
                } catch (Throwable t) {
                    System.out.println("[JDBC] unable to unregister driver " + d.getClass() + ": " + t);
                }
            });
```

eolivelli:
okay, as discussed off line

PR #237, opened 2018-08-02
Recalculate BLink node size if required

diegosalvi:
* Recalculate BLink node size if required
* Require recalculation if metadata version = 0

### DIFF
--- a/herddb-core/src/main/java/herddb/index/blink/BLinkKeyToPageIndex.java
+++ b/herddb-core/src/main/java/herddb/index/blink/BLinkKeyToPageIndex.java
@@ -371,6 +371,8 @@ public class BLinkKeyToPageIndex implements KeyToPageIndex {
         /**
          * Added flags options after version in metadata and forced byte size recalculation due to changes
          * on size evaluation algorithm
+         *
+         * @since 0.6.0
          */
         private static final long VERSION_1 = 1L;
 
--- a/herddb-core/src/main/java/herddb/index/blink/BLinkKeyToPageIndex.java
+++ b/herddb-core/src/main/java/herddb/index/blink/BLinkKeyToPageIndex.java
@@ -100,7 +100,7 @@ public class BLinkKeyToPageIndex implements KeyToPageIndex {
     public static String deriveIndexName(String tableName) {
         return tableName + "_primary";
     }
-    
+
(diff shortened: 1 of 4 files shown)


herddb-core/src/main/java/herddb/index/blink/BLinkKeyToPageIndex.java
```diff
@@ -426,9 +429,18 @@ public long evaluateAll(Bytes key, Long value) {
...
                 ExtendedDataInputStream edis = new ExtendedDataInputStream(bis)) {
 
-                /* flags for future implementations, actually unused */
+                long version = edis.readVLong();
+
+                /*
+                 * Check if byte size needs to be recalculated (between v.0 and v.1 was changed size evaluation
+                 * algorithm so v.0 stored size is meaningless)
+                 */
+                boolean recalculateSize = version == 0;
```
eolivelli: Constant for version 0 ?

herddb-core/src/main/java/herddb/index/blink/BLinkKeyToPageIndex.java
```diff
@@ -426,9 +429,18 @@ public long evaluateAll(Bytes key, Long value) {
...
+
+                /* flags for future implementations, actually unused (exists from version 1)*/
                 @SuppressWarnings("unused")
-                long flags = edis.readVLong();
+                long flags = version > 0 ? edis.readVLong() : 0L;
```
eolivelli: Constant ?

herddb-utils/src/main/java/herddb/index/blink/BLinkMetadata.java
```diff
@@ -90,6 +90,7 @@ public String nodesToStrings() {
     public static final class BLinkNodeMetadata<K> {
 
         public static final long NO_LINK = -1;
+        public static final long CURRENT_VERSION = 1L;
```
eolivelli: Add constant for v0 and v1.
Maybe add @ since 0.6.0 in v1

herddb-core/src/main/java/herddb/index/blink/BLinkKeyToPageIndex.java
diegosalvi: +1 for no flag constant

diegosalvi:
Backward compatibility test is some lacking. I've added a test to check unknown_size correct behaviour (which is half of the compatibility part)

PR #241, opened 2018-08-07
Run fsync in a separate thread

eolivelli:
This will allow regular writers to move forward without blocking for fsync.

Restore herddb.file.maxsynctime default to 1ms
Drop some lines of log during checkpoints
Introduce new system property herddb.file.maxconcurrentfsyncs=1
Usually HerdDB runs in systems with a single disk, it is not very useful to perform multiple fsyncs concurrently.

herddb.file.maxconcurrentfsyncs deals only with the FileCommitLog, it is not about index/data page writes

### DIFF
--- a/herddb-core/src/main/java/herddb/file/FileCommitLog.java
+++ b/herddb-core/src/main/java/herddb/file/FileCommitLog.java
@@ -50,6 +50,7 @@ import herddb.utils.ExtendedDataOutputStream;
 import herddb.utils.FileUtils;
 import herddb.utils.SimpleBufferedOutputStream;
 import herddb.utils.SystemProperties;
+import java.util.concurrent.ExecutorService;
 import org.apache.bookkeeper.stats.Counter;
 import org.apache.bookkeeper.stats.OpStatsLogger;
 import org.apache.bookkeeper.stats.StatsLogger;
@@ -78,6 +79,7 @@ public class FileCommitLog extends CommitLog {
     private final OpStatsLogger statsFsyncTime;
     private final OpStatsLogger statsEntryLatency;
     private final Counter queueSize;
+    private final ExecutorService fsyncThreadPool;
 
     private final static int WRITE_QUEUE_SIZE = SystemProperties.getIntSystemProperty(
             "herddb.file.writequeuesize", 10_000_000);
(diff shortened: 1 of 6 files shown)


herddb-core/src/main/java/herddb/file/FileCommitLog.java
```diff
@@ -142,10 +152,20 @@ public void sync() throws IOException {
         @Override
         public void close() throws LogNotAvailableException {
             try {
-                out.close();
-                channel.close();
-            } catch (IOException err) {
-                throw new LogNotAvailableException(err);
+
+                try {
+                    out.flush();
```
eolivelli: ensure fsync at close of a file

herddb-core/src/main/java/herddb/file/FileCommitLog.java
```diff
@@ -272,24 +320,20 @@ public void run() {
...
                         if (!doneEntries.isEmpty()) {
-                            synch();
-                            for (LogEntryHolderFuture e : doneEntries) {
-                                if (e.sync) {
-                                    statsEntryLatency.registerSuccessfulEvent(now - e.entry.timestamp, TimeUnit.MILLISECONDS);
-                                    e.syncDone();
-                                }
-                            }
-                            doneEntries.clear();
+                            flush();
```
eolivelli: only "flush" in this thread

herddb-core/src/test/java/herddb/file/FileCommitLogTest.java
```diff
@@ -180,59 +184,63 @@ public void accept(LogSequenceNumber t, LogEntry u) {
...
-            }
-        }
-        final long _endWrite = System.currentTimeMillis();
-        AtomicInteger readCount = new AtomicInteger();
-        try (CommitLog log = manager.createCommitLog("tt", "aa", "nodeid");) {
-            log.recovery(LogSequenceNumber.START_OF_TIME, new BiConsumer<LogSequenceNumber, LogEntry>() {
-                @Override
-                public void accept(LogSequenceNumber t, LogEntry u) {
-                    readCount.incrementAndGet();
+        try (FileCommitLogManager manager = new FileCommitLogManager(folder.newFolder().toPath(), 64 * 1024 * 1024);) {
```
eolivelli: this change seems big, but it is only about wrapping new FileCommitLogManager in a try with resources

Issue #244, opened 2018-08-08
Error importing table from MySQL


eolivelli:
CREATE TABLE `ao_21d670_whitelist_rules` (
  `ALLOWINBOUND` tinyint(1) DEFAULT NULL,
  `EXPRESSION` longtext COLLATE utf8_bin NOT NULL,
  `ID` int(11) NOT NULL AUTO_INCREMENT,
  `TYPE` varchar(255) COLLATE utf8_bin NOT NULL,
  PRIMARY KEY (`ID`)
) 

this is the error.
error:java.sql.SQLException: herddb.model.StatementExecutionException: java.lang.IllegalArgumentException: Unexpected error on cast of value http://www.atlassian.com/* (class herddb.utils.RawString): java.lang.ClassCastException: herddb.utils.RawString cannot be cast to java.lang.Number  query was 'INSERT INTO `ao_21d670_whitelist_rules` VALUES (?,?,?,?)', with values [[0, http://www.atlassian.com/*, 3, WILDCARD_EXPRESSION], [0, 3d194026-dc63-385b-86a6-0cc89001d9a1, 4, APPLICATION_LINK], [0, null://null/*, 5, WILDCARD_EXPRESSION], [0, 33ba34ad-a52f-3e7f-8db1-14d253719ac1, 6, APPLICATION_LINK], [0, 3d194026-dc63-385b-86a6-0cc89001d9a1, 7, APPLICATION_LINK], [0, 33ba34ad-a52f-3e7f-8db1-14d253719ac1, 8, APPLICATION_LINK], [0, http://www.atlassian.com/*, 9, WILDCARD_EXPRESSION], [0, f6ce63b0-7606-3171-9b42-d087e3791629, 10, APPLICATION_LINK]]

eolivelli:
Stacktrace:
SEVERE: NettyChannel{name=unnamed, id=1, socket=[id: 0x9eff6bd3, L:/127.0.0.1:7000 - R:/127.0.0.1:59214] pending 0 msgs}: error java.lang.RuntimeException: Unexpected error on cast of value http://www.atlassian.com/* (class herddb.utils.RawString): java.lang.ClassCastException: herddb.utils.RawString cannot be cast to java.lang.Number
java.lang.RuntimeException: Unexpected error on cast of value http://www.atlassian.com/* (class herddb.utils.RawString): java.lang.ClassCastException: herddb.utils.RawString cannot be cast to java.lang.Number
[...]

PR #245, opened 2018-08-08
Add 'sysstatements' and provide better error reporting to the client in case of bad query

eolivelli:


herddb-core/src/main/java/herddb/sql/SQLRecordFunction.java
```diff
@@ -60,20 +60,24 @@ public SQLRecordFunction(List<String> columns, Table table, List<CompiledSQLExpr
...
+                }
+                columnName = column.name;
+                Object value = RecordSerializer.convert(column.type, e.evaluate(bean, context));
+                res.put(columnName, value);
             }
-            columnName = column.name;
-            Object value = RecordSerializer.convert(column.type, e.evaluate(bean, context));
-            res.put(columnName, value);
+            return RecordSerializer.toRecord(res, table).value.data;
+        } catch (IllegalArgumentException err) {
```
eolivelli: This is to catch bad errors in INSERTS, see #244

PR #248, opened 2018-08-09
BLink clanups and less copies on lead load

diegosalvi:
* Removed leftover Comparable and casts on BLink
* Skip data map copies on leaf load


eolivelli:
+1
is spotbugs happy with this change ?

PR #251, opened 2018-08-30
Do not decode query results on Netty thread

eolivelli:


eolivelli:
cc @aluccaroni

herddb-core/src/main/java/herddb/client/ScanResultSet.java
```diff
@@ -49,9 +50,10 @@ public void close() {
     public List< Map<String, Object>> consume() throws HDBException {
```
eolivelli: this method is used for tests only

PR #254, opened 2018-09-13
Save a Bytes allocation on the read path

eolivelli:


herddb-core/src/main/java/herddb/sql/SQLRecordKeyFunction.java
```diff
@@ -133,7 +133,7 @@ public boolean isFullPrimaryKey() {
         }
         try {
             // maybe this is only a partial primary key
-            byte[] result = RecordSerializer.serializePrimaryKey(pk, table, pkColumnNames).data;
+            byte[] result = RecordSerializer.serializePrimaryKeyRaw(pk, table, pkColumnNames);
```
eolivelli: this is actually the fix

ycsb-runner/setsenv_bench_jdk11.sh
```diff
@@ -0,0 +1,17 @@
+# Basic Environment and Java variables
+
+JAVA_HOME=~/dev/jdk-11
```
eolivelli: this is unrelated to the change I can add to another PR

PR #266, opened 2018-09-29
Use correctly RelCollectionTraitDef

eolivelli:
- Use correctly RelCollectionTraitDef
- Fix test on NATURAL JOIN (the test was wrong)

Thanks to @zabetak for very useful suggestions


zabetak:
Looks better now. I'm glad that it was just this :)

PR #270, opened 2018-10-02
Fix BIGINT conversion

diegosalvi:
When inserting a big enough long/bingint value through jdbc the value lost "precision" due to internal conversion to double value.

### DIFF
--- a/herddb-jdbc/src/test/java/herddb/jdbc/CheckBigIntConversion.java
+++ b/herddb-jdbc/src/test/java/herddb/jdbc/CheckBigIntConversion.java
@@ -1,0 +1,57 @@
+package herddb.jdbc;
+
+import java.sql.Connection;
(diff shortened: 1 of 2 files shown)


herddb-jdbc/src/test/java/herddb/jdbc/CheckBigIntConversion.java
```diff
@@ -0,0 +1,57 @@
...
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+import herddb.client.ClientConfiguration;
+import herddb.client.HDBClient;
+import herddb.server.Server;
+import herddb.server.ServerConfiguration;
+import herddb.server.StaticClientSideMetadataProvider;
+
+public class CheckBigIntConversion {
```
eolivelli: typo: CheckBigIntConversionTest ?

PR #278, opened 2018-10-15
Do not block ServerSideConnectionPeer threads on writes to log

eolivelli:
This is a huge refactor to enable DML operations to not block ServerSideConnection peer, that is the main threadpool which handles incoming requests.
The idea is that DML operation must wait for the log to write, so ideally it is better to suspend execution, leaving the CPU to other requests, and then continue the execution after the log acknowledges the write (mostly the fsync is the problem).

Most of the work is a big refactor to use CompletableFuture inside DBManager, TableSpaceManager and TableManager.
The main lock in TableSpaceManager has been converted to StampedLock, because now the thread which releases the lock (the post log write execution) is not the thread which requests it (usually ServerSideConnectionPeer threads).

A bunch on improvements around SQLPlannerOperationStatement has been done to skip a useless step in case of very simple DML statements (SimpleUpdateOp, SimpleDeleteOp, SimpleInsertOp), the same for basic SELECT operations (every SQLPlannerOperationStatement which directly wraps a single ScanStatement)

The patch does not change tests, it adds some test cases. A follow up work would to add tests around unlocking of keys after "failed" DMLs (this is mostly uncovered, or it is not covered explicitly)

There are changes around FileCommitLog and BookKeeperCommitLog, because we have to return a CompletableFuture, to be the hook for continuing the DML work.

A big problem is about 'completed' CompletableFutures, because "thenApply" and other methods of CompletableFuture has a weird behaviour if you attach the listener after the completion of the future: Exceptions will be wrapped with CompletionException in a quite unpredictable way.

Many parts of old SQL Planner were dropped because they should have refactored to the new way, that would be useless.


herddb-core/src/main/java/herddb/model/ExecutionPlan.java
```diff
@@ -37,67 +35,33 @@
...
-        List<InsertStatement> insertStatements) {
+            Aggregator mainAggregator,
+            ScanLimits limits,
+            TupleComparator comparator,
+            PlannerOp originalRoot) {
         this.mainStatement = mainStatement;
         this.mainAggregator = mainAggregator;
         this.limits = limits;
         this.comparator = comparator;
-        this.dataSource = dataSource;
```
eolivelli: this was for legacy planner.....dropped...

herddb-core/src/main/java/herddb/model/StatementEvaluationContext.java
```diff
@@ -89,4 +91,20 @@ public Object getJdbcParameter(int index) throws StatementExecutionException {
...
+    public long getTableSpaceLock() {
+        return tableSpaceLock;
+    }
+
+    public void setTableSpaceLock(long tableSpaceLock) {
+        this.tableSpaceLock = tableSpaceLock;
+    }
+
+    public boolean isTableSpaceLockWriteLockProtomoted() {
+        return tableSpaceLockWriteLockProtomoted;
```
eolivelli: useless, to be removed

herddb-core/src/main/java/herddb/model/planner/BindableTableScanOp.java
```diff
@@ -19,54 +19,17 @@
...
 
 /**
  * TableScanOp + ProjectOp + FilerOp
  *
  * @author eolivelli
  */
-public class BindableTableScanOp implements PlannerOp {
-
-    final ScanStatement statement;
+public class BindableTableScanOp extends SimpleScanOp {
```
eolivelli: this is the first of a bunch of semplifications, a bunch of classes had the same implementation, which is not SimpleScanOp

PR #282, opened 2018-10-23
Do not allocate HashMaps during updates and inserts

eolivelli:
Do not create and fill temporary HashMaps when executing updates.

The new record to be built can simply be re-created on the flight without materializing all the values and filling 2 hashmaps.

### DIFF
--- a/herddb-core/src/main/java/herddb/codec/RecordSerializer.java
+++ b/herddb-core/src/main/java/herddb/codec/RecordSerializer.java
@@ -26,7 +26,6 @@ import herddb.model.ColumnTypes;
 import herddb.model.Record;
 import herddb.model.Table;
 import herddb.utils.Bytes;
-import herddb.utils.ExtendedDataInputStream;
 import herddb.utils.ExtendedDataOutputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
@@ -50,6 +49,7 @@ import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
 import java.util.function.BiConsumer;
+import java.util.function.Function;
 
 /**
  * Record conversion to byte[]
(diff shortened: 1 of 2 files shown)


herddb-core/src/main/java/herddb/sql/SQLRecordFunction.java
```diff
@@ -46,37 +47,81 @@
...
+        }
     }
 
     @Override
     public byte[] computeNewValue(Record previous, StatementEvaluationContext context, TableContext tableContext) throws StatementExecutionException {
         try {
-            Map<String, Object> res = previous != null ? new HashMap<>(previous.toBean(table)) : new HashMap<>(table.columns.length);
+            byte[] oldRes = _computeNewValueOld(previous, context);
+            byte[] newRes = _computeNewValue(previous, context, tableContext);
+            if (!Arrays.equals(oldRes, newRes)) {
```
diegosalvi: Computing both of them don't we loose the patch benefits?
It should be under some assertion or debug mode to avoid computations in standard runtime

eolivelli: it is only debug during tests, to be sure that the new algorithm serializes data exactly as the previous one

eolivelli:
all tests are passing. I will clean up the code.

I have another patch in the stack around this change, let's do one change at a time.

I will perform profiling after this change

herddb-core/src/main/java/herddb/codec/RecordSerializer.java
```diff
@@ -672,11 +674,11 @@ public static Object deserializePrimaryKey(byte[] key, Table table) {
...
 
     public static Bytes serializeValue(Map<String, Object> record, Table table) {
-        return new Bytes(serializeValueRaw(record, table));
+        return new Bytes(serializeValueRaw(record, table, 0));
     }
-        
-    public static byte[] serializeValueRaw(Map<String, Object> record, Table table) {
-        ByteArrayOutputStream value = new ByteArrayOutputStream(INITIAL_BUFFER_SIZE);
+
+    public static byte[] serializeValueRaw(Map<String, Object> record, Table table, int expectedSize) {
```
eolivelli: I expect this to be another subtle optimization during updates, in fact we will reallocate a buffer of the same size of the previous version of the record, so we want fall into reallocations and copies due to buffer expansions.

In the future for INSERT we could have an average record size per table

eolivelli:
@diegosalvi  not this is good to go

PR #298, opened 2018-11-19
Add ability to flush log counting pending bytes and not operations

eolivelli:
Make configurable every parameter of FileCommitLog and add a max number of bytes waiting for flush.
This is better that putting a limit on the number of operations, because this kind of configuration fits better the underlying disk/SO and it does not depend on the size of writes (1000 writes of 1KB are different than 1000 writes of 1MB)

In order to add tests new metrics have been added as well

### DIFF
--- a/herddb-core/src/main/java/herddb/file/FileCommitLog.java
+++ b/herddb-core/src/main/java/herddb/file/FileCommitLog.java
@@ -49,6 +49,8 @@ import herddb.utils.ExtendedDataOutputStream;
 import herddb.utils.FileUtils;
 import herddb.utils.SimpleBufferedOutputStream;
 import herddb.utils.SystemProperties;
+import io.netty.buffer.ByteBuf;
+import io.netty.util.ByteProcessor;
 import java.nio.channels.ClosedChannelException;
 import java.util.Collections;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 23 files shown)


eolivelli:
CI results:
https://dev.majordodo.org/jenkins/job/HerdDB/1401/

herddb-core/src/main/java/herddb/file/FileCommitLog.java
```diff
@@ -364,20 +374,23 @@ public void run() {
...
                     boolean timedOut = false;
                     if (entry != null) {
                         batch.add(entry);
                         count++;
+                        unwrittenBytes += entry.entry.length;
                     } else {
                         timedOut = true;
                     }
-                    if (timedOut || count >= MAX_UNSYNCHED_BATCH) {
+                    if (timedOut || count >= MAX_UNSYNCHED_BATCH || unwrittenBytes >= MAX_UNSYNCHED_BATCH_BYTES) {
```
eolivelli: this is the actual impl

herddb-core/src/main/java/herddb/file/FileCommitLog.java
```diff
@@ -141,13 +147,16 @@ private CommitFileWriter(long ledgerId, long sequenceNumber) throws IOException
             writtenBytes = 0;
         }
 
-        public void writeEntry(long seqnumber, LogEntry edit) throws IOException {
+        public int writeEntry(long seqnumber, byte[] entry) throws IOException {
 
             this.out.writeByte(ENTRY_START);
             this.out.writeLong(seqnumber);
-            int written = edit.serialize(this.out);
```
eolivelli: entry now is serialized not in spool thread, but in the called thread.
I can't use ByteBuf because we have ExtendedDataOutputStream, which is not a "Channel" and it cannot deal with NIO Buffers (a mem copy is needed).

eolivelli:
@diegosalvi added testcases

build results:
https://dev.majordodo.org/jenkins/job/HerdDB/1404/

PR #301, opened 2018-11-23
Introduce PreparedStatementAsyc and speed up MySQL imports

eolivelli:
This change introduces an new API, PreparedStatementAsyc which is an extension to JDBC PreparedStatement.

The first goal is to speed up MySQL import of huge databases, by sending parallel writes to the server, but the API will be of general use.


eolivelli:
There is still a problem on the MySQL importer facility, it does not deal with transactions and multiple concurrent updates. We have to create a new SQL Connection for each statement, otherwise the "current transaction is" bound to the java.sql.Connection is in an undefined state

eolivelli:
@diegosalvi  now the patch is ready for review.
After mergin this one I will update the wiki with the 'fast MySQL import"

# scan the file for DDL (create table...)
bin/herddb-cli.sh --mysql -f backup.sql -filter ddl

# execute DML statements (Inserts....) in "async" mode
bin/herddb-cli.sh --mysql -f backup.sql -filter dml -a

using "async" mode the client continues to send data without blocking. In the future the procedure can be enhanced, but with this change is enough to boost x10 the import procedure

herddb-core/src/main/java/herddb/client/HDBConnection.java
```diff
@@ -186,14 +194,83 @@ public DMLResult executeUpdate(String tableSpace, String query, long tx, boolean
...
+            return FutureUtils.exception(new HDBException("client is closed"));
+        }
+        CompletableFuture<DMLResult> res = new CompletableFuture<>();
+
+        AtomicInteger count = new AtomicInteger(10);
+        executeStatementAsyncInternal(tableSpace, res, query, tx, returnValues, usePreparedStatement, params, count);
+        return res;
+    }
+
+    private void executeStatementAsyncInternal(String tableSpace, CompletableFuture<DMLResult> res, String query, long tx, boolean returnValues, boolean usePreparedStatement, List<Object> params, AtomicInteger count) {
```
diegosalvi: Wouldn't be cleaner to return a CompletableFuture and not require it as an input?

eolivelli: No, it is like a recursion...

PR #302, opened 2018-11-23
Use posix_fallocate to preallocate space on disk for FileCommitLog

eolivelli:


eolivelli:
We do not have enough evidence of performance improvements, let's come back to this implementation in the future

PR #304, opened 2018-11-30
Add unfolded var int read and write

diegosalvi:
Add an unfolded (no loop) version of write and read var int

### DIFF
--- a/herddb-utils/src/main/java/herddb/utils/ExtendedDataInputStream.java
+++ b/herddb-utils/src/main/java/herddb/utils/ExtendedDataInputStream.java
@@ -29,8 +29,13 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
  * Extended version of DataInputStream
  *
  * @author enrico.olivelli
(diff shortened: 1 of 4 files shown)


eolivelli:
@diegosalvi do you have numbers to share ? Is there a real benefit ?

I will run YCSB with @hamadodene

herddb-utils/src/main/java/herddb/utils/ExtendedDataInputStream.java
```diff
@@ -29,9 +29,14 @@
  * Extended version of DataInputStream
  *
  * @author enrico.olivelli
+ * @author diego.salvi
+ *
+ * @see ExtendedDataOutputStream
  */
 public class ExtendedDataInputStream extends DataInputStream {
 
+    private static final boolean USE_FOLDED_VAR_INT = Boolean.getBoolean("herddb.vint.read.folded");
```
eolivelli: can we use SystemProperties.getBooleanSystemProperty and set the default to "true" ?

PR #307, opened 2018-12-03
Remove empty syncs and write data as soon as possible

diegosalvi:
Removed massive empty syncs on FileCommitLog and changed flush and sync logic to write as soon as possible but sync only when enough unsynced data is reached.

Added entrySyncLatency stat and renamed batchWriteSize and batchWriteBytes to syncBatchSize and syncBatchBytes


eolivelli:
great !

PR #308, opened 2018-12-04
Do not create temporary byte[] for entry serializations ni FileCommitLog

eolivelli:
During the development on 0.8.0 we moved serialization of entries a "logEntry", but this implicitly materializes the entry as byte[].
The old behaviour was to serialize the contents of the entry to the OutputStream so without this additional memory copy.
I found this problem using JVM and TLAB allocations.

### DIFF
--- a/herddb-core/src/main/java/herddb/file/FileCommitLog.java
+++ b/herddb-core/src/main/java/herddb/file/FileCommitLog.java
@@ -146,11 +146,10 @@ public class FileCommitLog extends CommitLog {
             writtenBytes = 0;
         }
 
-        private int writeEntry(long seqnumber, byte[] entry) throws IOException {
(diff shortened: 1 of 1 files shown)


herddb-core/src/main/java/herddb/file/FileCommitLog.java
```diff
@@ -406,7 +405,7 @@ public void run() {
                 long unsyncedBytes = 0;
                 int unsyncedCount = 0;
                 while (!closed || !writeQueue.isEmpty()) {
-                    LogEntryHolderFuture entry = writeQueue.poll(maxSyncTime, TimeUnit.MILLISECONDS);
+                    LogEntryHolderFuture entry = writeQueue.poll(maxSyncTime, TimeUnit.MICROSECONDS);
```
eolivelli: @diegosalvi  this is a problem. I have to revert this line before committing

PR #311, opened 2018-12-07
Open FileCommitLog with O_DIRECT on jdk10+

eolivelli:


herddb-utils/src/test/java/herddb/utils/ODirectFileOutputStreamTest.java
```diff
@@ -0,0 +1,220 @@
...
+import static org.junit.Assert.assertTrue;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+/**
+ *
+ * @author enrico.olivelli
+ */
+public class ODirectFileOutputStreamTest {
```
eolivelli: This test should be moved to core, because mukti release jars are not picked up with tests in the same project, java needs a real jar and not target/classes

diegosalvi:
Great work!!!!

PR #312, opened 2018-12-11
Enable O_DIRECT for index and datapages

diegosalvi:
Added O_DIRECT support for read/write data and index pagest to avoid double caches and disc cache pollution


herddb-core/src/main/java/herddb/file/FileDataStorageManager.java
```diff
@@ -236,14 +259,40 @@ private boolean isTableOrIndexCheckpointsFile(Path path) {
...
+            }
+            int numRecords = dataIn.readInt();
+            result = new ArrayList<>(numRecords);
+            for (int i = 0; i < numRecords; i++) {
+                byte[] key = dataIn.readArray();
+                byte[] value = dataIn.readArray();
+                result.add(new Record(new Bytes(key), new Bytes(value)));
+            }
+            hashFromDigest = hash.hash();
+            hashFromFile = dataIn.readLong();
```
eolivelli: Can we add a comment about padding?
The fille will be padded with zeroes

herddb-core/src/test/java/herddb/utils/ODirectFileInputStreamTest.java
```diff
@@ -0,0 +1,162 @@
...
+import java.nio.file.Path;
+import java.util.Random;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+/**
+ * Testing O_DIRECT. This file should be in core project, because it leverages Multi-Release JAR
+ * feature of JDK9
```
eolivelli: Jdk10

herddb-utils/src/main/java/herddb/utils/ODirectFileOutputStream.java
```diff
@@ -35,22 +36,41 @@
...
     final int alignment;
+    final int batchSize;
     int writtenBlocks;
 
     public ODirectFileOutputStream(Path p) throws IOException {
+        this(p,1);
+    }
+
+    private static final OpenOption[] DEFAULT_OPTIONS = new OpenOption[] { StandardOpenOption.CREATE, StandardOpenOption.WRITE };
+    public ODirectFileOutputStream(Path p, int btachBlocks, OpenOption... options) throws IOException {
```
eolivelli: Typo

herddb-core/src/main/java/herddb/file/FileDataStorageManager.java
```diff
@@ -100,20 +107,27 @@
      */
     public static final int COPY_BUFFERS_SIZE = 64 * 1024;
 
+    /** Standard blocks batch number for o_direct procedures */
+    public static final int O_DIRECT_BLOCK_BATCH = 16;
```
eolivelli: System property?

herddb-core/src/main/java/herddb/server/ServerConfiguration.java
```diff
@@ -68,7 +69,7 @@
 
     public final static String PROPERTY_DEFERRED_SYNC_PERIOD = "txlog.deferredsyncperiod";
     public final static int PROPERTY_DEFERRED_SYNC_PERIOD_DEFAULT = 0;  /* disabled */
-    
+  
     public final static String PROPERTY_TXLOG_USE_ODIRECT = "txlog.use_o_direct";
     public final static boolean PROPERTY_TXLOG_USE_ODIRECT_DEFAULT = false;
```
eolivelli: Can we add a system property for this default and the others?
Just a 'herddb.file.use_o_direct_default'
To be used for all of the 3 params.

I would like to setup a CI job which runs all the suite with o_direct=true

This way we can cover better all of the usages

herddb-core/src/main/java/herddb/file/FileDataStorageManager.java
diegosalvi: sure we can add some comment... padding itself isn't a problem here because the reader already known how much data need to read and stops without reaching padding or eof (if the file isn't truncated obviously)

eolivelli: Some time ago I was trying to change the read procedure in order to look for the checksum at the end of the file.
But it was not possible because we are already rewriting pages and we have already churn at the end of data pages

eolivelli:
@diegosalvi  travis is failing
https://travis-ci.org/diennea/herddb/builds/466933368?utm_source=github_status&utm_medium=notification

herddb-core/src/main/java/herddb/server/ServerConfiguration.java
diegosalvi: Added a system property herddb.file.use_o_direct_default to handle all 3 defaults

eolivelli:
@diegosalvi  travis in not happy about your patch

https://travis-ci.org/diennea/herddb/builds/466944542?utm_source=github_status&utm_medium=notification

herddb-core/src/main/java/herddb/file/FileDataStorageManager.java
diegosalvi: Added a comment about while padding doesn't matters

PR #313, opened 2018-12-14
Get FileStore block size after file creation in O_DIRECT mode

diegosalvi:


herddb-utils/src/main/java/herddb/utils/ODirectFileInputStream.java
```diff
@@ -50,12 +50,13 @@ public ODirectFileInputStream(Path path) throws IOException {
     }
 
     public ODirectFileInputStream(Path path, int batchBlocks) throws IOException {
-        this.batchBlocks = batchBlocks;
 
+        fc = OpenFileUtils.openFileChannelWithO_DIRECT(path, StandardOpenOption.READ);
```
eolivelli: if we have an error during getBlockSize we will have a dandling open file

PR #324, opened 2019-01-16
Cache Pattern for Like operator

eolivelli:
Cache the "Pattern" in LIKE operator in case of a constant expression,
this way we will "parse" the pattern only once in case of repeated queries.

Follow up for  789e436f0b29be184d13d2d81a8128b3f8424d32

### DIFF
--- a/herddb-core/src/test/java/herddb/core/SimpleOperatorsTest.java
+++ b/herddb-core/src/test/java/herddb/core/SimpleOperatorsTest.java
@@ -287,6 +287,13 @@ public class SimpleOperatorsTest {
             try (DataScanner scan1 = scan(manager, "SELECT * FROM tblspace1.tsql WHERE 'AbBbCc' LIKE '%AbBbCc%'", Collections.emptyList());) {
                 assertEquals(1, scan1.consume().size());
             }
+            
+            // bad "constant" pattern
+            herddb.utils.TestUtils.assertThrows(herddb.core.HerdDBInternalException.class, () -> {
+                scan(manager, "SELECT * FROM tblspace1.tsql WHERE 'AbBbCc' LIKE '%AbBbC(.c%'", Collections.emptyList());
+            });
+            
+            
             try (DataScanner scan1 = scan(manager, "SELECT * FROM tblspace1.tsql WHERE 'AbBbCc' LIKE ?", Arrays.asList("%AbBbCc%"));) {
                 assertEquals(1, scan1.consume().size());
             }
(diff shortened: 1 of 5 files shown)


eolivelli:
@diegosalvi  I have committed a failing test.
We have to implement a better conversion from LIKE to Java RegExps, otherwise the user will find infinite odd cases

diegosalvi:
I have added a simple parsing of incoming string: regex pattern will be totally quoted with just some "unquote, pattern, quote again" sequences on % and _ (`\\E.*\\Q` and `\\E.?\\Q`).

herddb-core/src/test/java/herddb/core/SimpleOperatorsTest.java
```diff
@@ -287,6 +287,16 @@ public void simpleArithmeticOperationsTest() throws Exception {
             try (DataScanner scan1 = scan(manager, "SELECT * FROM tblspace1.tsql WHERE 'AbBbCc' LIKE '%AbBbCc%'", Collections.emptyList());) {
                 assertEquals(1, scan1.consume().size());
             }
+            
+            // bad "constant" pattern
+            herddb.utils.TestUtils.assertThrows(herddb.core.HerdDBInternalException.class, () -> {
+                scan(manager, "SELECT * FROM tblspace1.tsql WHERE 'AbBbCc' LIKE '%AbBbC(.c%'", Collections.emptyList());
```
eolivelli: this part is to be dropped @diegosalvi 
it would be awesome to add a test with a wrong pattern, but I guess there is no real "bad" patten for LIKE sintax

herddb-utils/src/test/java/herddb/utils/SQLRecordPredicateFunctionsTest.java
```diff
@@ -41,6 +41,13 @@ public void testCompareAndLike() throws Exception {
         assertTrue(SQLRecordPredicateFunctions.like("test", "%es%"));
         assertFalse(SQLRecordPredicateFunctions.like("tesst", "te_t"));
         assertTrue(SQLRecordPredicateFunctions.like("test", "te_t"));
+        assertTrue(SQLRecordPredicateFunctions.like("bar (foo)", "%(foo)%"));
```
eolivelli: please also add a test without any '%' and '_'

it may sound weird but it is quite common in generated SQL

diegosalvi:
Added some testcases.
There was an error on `_` pattern that had to be converted to `.{1} `not `.?`

diegosalvi:
Actually LIKE doesn't support an escape character

eolivelli:
We can merge this one after Travis @diegosalvi 
Thank you for having picked up the work

PR #327, opened 2019-01-27
Allow Bytes to wrap a slice of byte[]

eolivelli:
- Encapsulate Bytes internal array, drop the only public constructor
- Let Bytes wrap a generic slice of a byte[], adding *offset* and *length* fields
- Use Bytes whenever possible, save resources by not creating temporary byte[] and saving memory copies on the hottest paths
- Prepare the code to be able to load data pages as a single byte[] and have key/values as slices on that buffer (this will boost loading of pages and save GC cycles)

### DIFF
--- a/herddb-core/src/main/java/herddb/codec/RecordSerializer.java
+++ b/herddb-core/src/main/java/herddb/codec/RecordSerializer.java
@@ -60,6 +60,29 @@ public final class RecordSerializer {
 
     private static final int INITIAL_BUFFER_SIZE = SystemProperties.getIntSystemProperty("herddb.serializer.initbufsize", 4 * 1024);
 
+    public static Object deserialize(Bytes data, int type) {
+        switch (type) {
+            case ColumnTypes.BYTEARRAY:
+                return data.to_array();
+            case ColumnTypes.INTEGER:
+                return data.to_int();
+            case ColumnTypes.LONG:
+                return data.to_long();
+            case ColumnTypes.STRING:
+                return data.to_RawString();
+            case ColumnTypes.TIMESTAMP:
+                return data.to_timestamp();
+            case ColumnTypes.NULL:
+                return null;
+            case ColumnTypes.BOOLEAN:
+                return data.to_boolean();
+            case ColumnTypes.DOUBLE:
+                return data.to_double();
+            default:
+                throw new IllegalArgumentException("bad column type " + type);
+        }
+    }
+    
     public static Object deserialize(byte[] data, int type) {
         switch (type) {
             case ColumnTypes.BYTEARRAY:
@@ -276,7 +299,7 @@ public final class RecordSerializer {
 
     public static void serializeTo(Object v, int type, ExtendedDataOutputStream out) throws IOException {
         if (v == null) {
-            out.writeArray(null);
+            out.writeNullArray();
             return;
         }
         switch (type) {
@@ -515,7 +538,7 @@ public final class RecordSerializer {
         if (table.getColumn(property) == null) {
             throw new herddb.utils.IllegalDataAccessException("table " + table.tablespace + "." + table.name + " does not define column " + property);
         }
-        try (ByteArrayCursor din = ByteArrayCursor.wrap(value.data);) {
+        try (ByteArrayCursor din = value.newCursor();) {
             while (!din.isEof()) {
                 int serialPosition;
                 serialPosition = din.readVIntNoEOFException();
@@ -536,7 +559,7 @@ public final class RecordSerializer {
 
     static Object accessRawDataFromValue(int index, Bytes value, Table table) throws IOException {
         Column column = table.getColumn(index);
-        try (ByteArrayCursor din = ByteArrayCursor.wrap(value.data);) {
+        try (ByteArrayCursor din = value.newCursor();) {
             while (!din.isEof()) {
                 int serialPosition;
                 serialPosition = din.readVIntNoEOFException();
@@ -557,7 +580,7 @@ public final class RecordSerializer {
 
     static int compareRawDataFromValue(int index, Bytes value, Table table, Object cvalue) throws IOException {
         Column column = table.getColumn(index);
-        try (ByteArrayCursor din = ByteArrayCursor.wrap(value.data);) {
+        try (ByteArrayCursor din = value.newCursor()) {
             while (!din.isEof()) {
                 int serialPosition;
                 serialPosition = din.readVIntNoEOFException();
@@ -578,9 +601,9 @@ public final class RecordSerializer {
 
     static Object accessRawDataFromPrimaryKey(String property, Bytes key, Table table) throws IOException {
         if (table.primaryKey.length == 1) {
-            return deserialize(key.data, table.getColumn(property).type);
+            return deserialize(key, table.getColumn(property).type);
         } else {
-            try (ByteArrayCursor din = ByteArrayCursor.wrap(key.data);) {
+            try (ByteArrayCursor din = key.newCursor();) {
                 for (String primaryKeyColumn : table.primaryKey) {
                     byte[] value = din.readArray();
                     if (primaryKeyColumn.equals(property)) {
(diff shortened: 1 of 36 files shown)


herddb-core/src/main/java/herddb/codec/RecordSerializer.java
```diff
@@ -60,6 +60,29 @@
 
     private static final int INITIAL_BUFFER_SIZE = SystemProperties.getIntSystemProperty("herddb.serializer.initbufsize", 4 * 1024);
 
+    public static Object deserialize(Bytes data, int type) {
```
eolivelli: this is a copy of the other serialize(byte[]), it is worth to cut and paste, in order to be able to skip copies

herddb-core/src/main/java/herddb/codec/RecordSerializer.java
```diff
@@ -60,6 +60,29 @@
...
+        switch (type) {
+            case ColumnTypes.BYTEARRAY:
+                return data.to_array();
+            case ColumnTypes.INTEGER:
+                return data.to_int();
+            case ColumnTypes.LONG:
+                return data.to_long();
+            case ColumnTypes.STRING:
+                return data.to_RawString();
```
eolivelli: this is zero-copy

herddb-core/src/main/java/herddb/codec/RecordSerializer.java
```diff
@@ -120,6 +143,50 @@ public static int deserializeCompare(byte[] data, int type, Object cvalue) {
                 throw new IllegalArgumentException("bad column type " + type);
         }
     }
+    
+    public static int deserializeCompare(Bytes data, int type, Object cvalue) {
```
eolivelli: same as above

herddb-core/src/main/java/herddb/codec/RecordSerializer.java
```diff
@@ -120,6 +143,50 @@ public static int deserializeCompare(byte[] data, int type, Object cvalue) {
...
+        switch (type) {
+            case ColumnTypes.BYTEARRAY:
+                //TODO: not perform copy
```
eolivelli: we do not have any special code path for columns of type byte[], we can work on this topic in the future if needed

herddb-core/src/main/java/herddb/codec/RecordSerializer.java
```diff
@@ -578,11 +645,11 @@ static int compareRawDataFromValue(int index, Bytes value, Table table, Object c
...
     static Object accessRawDataFromPrimaryKey(String property, Bytes key, Table table) throws IOException {
         if (table.primaryKey.length == 1) {
-            return deserialize(key.data, table.getColumn(property).type);
+            return deserialize(key, table.getColumn(property).type);
         } else {
-            try (ByteArrayCursor din = ByteArrayCursor.wrap(key.data);) {
+            try (ByteArrayCursor din = key.newCursor();) {
                 for (String primaryKeyColumn : table.primaryKey) {
-                    byte[] value = din.readArray();
+                    Bytes value = din.readBytesNoCopy();
```
eolivelli: this is an improvement as it does not need to create a new array

herddb-core/src/main/java/herddb/codec/RecordSerializer.java
```diff
@@ -595,12 +662,12 @@ static Object accessRawDataFromPrimaryKey(String property, Bytes key, Table tabl
...
         if (table.primaryKey.length == 1) {
-            return deserialize(key.data, column.type);
+            return deserialize(key, column.type);
         } else {
             final String cname = column.name;
-            try (ByteArrayCursor din = ByteArrayCursor.wrap(key.data);) {
+            try (ByteArrayCursor din = key.newCursor();) {
                 for (String primaryKeyColumn : table.primaryKey) {
-                    byte[] value = din.readArray();
+                    Bytes value = din.readBytesNoCopy();
```
eolivelli: this is an improvement as it does not need to create a new array

herddb-core/src/main/java/herddb/codec/RecordSerializer.java
```diff
@@ -613,12 +680,12 @@ static Object accessRawDataFromPrimaryKey(int index, Bytes key, Table table) thr
...
         if (table.primaryKey.length == 1) {
-            return deserializeCompare(key.data, column.type, cvalue);
+            return deserializeCompare(key, column.type, cvalue);
         } else {
             final String cname = column.name;
-            try (ByteArrayCursor din = ByteArrayCursor.wrap(key.data)) {
+            try (ByteArrayCursor din = key.newCursor()) {
                 for (String primaryKeyColumn : table.primaryKey) {
-                    byte[] value = din.readArray();
+                    Bytes value = din.readBytesNoCopy();
```
eolivelli: this is an improvement as it does not need to create a new array

herddb-utils/src/main/java/herddb/utils/Bytes.java
```diff
@@ -117,9 +121,26 @@ public static Bytes from_long(long value) {
...
+    public static Bytes from_nullable_array(byte[] data) {
+        if (data == null) {
+            return null;
+        }
+        return new Bytes(data);
+    }
 
     public byte[] to_array() {
-        return data;
+        if (offset == 0 && buffer.length == length) {
```
eolivelli: no copy for the common case of a Bytes which is wrapping a whole byte[]

herddb-utils/src/main/java/herddb/utils/Bytes.java
```diff
@@ -327,11 +366,14 @@ public int compareTo(Bytes o) {
...
             return -1;
         }
-        return CompareBytesUtils.compare(this.data, o.data);
+        return CompareBytesUtils.compare(buffer, offset, offset + length, 
+                o.buffer, o.offset, o.offset + o.length);
     }
 
-    public static boolean startsWith(byte[] left, int len, byte[] right) {
-        for (int i = 0, j = 0; i < left.length && j < right.length && i < len; i++, j++) {
+    public static boolean startsWith(byte[] left, int offset, int bufferlen, int max, byte[] right) {
```
eolivelli: this is a new method, maybe worth spending some more time on review

herddb-core/src/main/java/herddb/codec/RecordSerializer.java
```diff
@@ -276,7 +343,7 @@ public static void skipTypeAndValue(ByteArrayCursor dii) throws IOException {
 
     public static void serializeTo(Object v, int type, ExtendedDataOutputStream out) throws IOException {
         if (v == null) {
-            out.writeArray(null);
+            out.writeNullArray();
```
diegosalvi: +1

herddb-core/src/main/java/herddb/codec/RecordSerializer.java
```diff
@@ -720,7 +787,8 @@ public static Bytes serializePrimaryKey(DataAccessor record, ColumnsList table,
             } catch (IOException err) {
                 throw new RuntimeException(err);
             }
-            return new Bytes(key.toByteArray());
+            //TODO: no copy
+            return Bytes.from_array(key.toByteArray());
```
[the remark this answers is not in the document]
diegosalvi: Yes it should be benchmarked... internal capacity grows doubling it every time. Pheraphs we could check wasted size percentage unpon copy or no copy actions

herddb-core/src/main/java/herddb/index/ConcurrentMapKeyToPageIndex.java
```diff
@@ -73,11 +73,11 @@ public void put(Bytes key, Long currentPage) {
     }
 
     private void keyAdded(Bytes key) {
-        usedMemory.addAndGet(key.data.length + ENTRY_OVERHEAD);
+        usedMemory.addAndGet(key.getLength() + ENTRY_OVERHEAD);
```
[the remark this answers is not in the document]
diegosalvi: Map entries are untouched, no need to change ENTRY_OVERHEAD.  Bytes estimatedSize should be reevaluated (pheraps)

PR #328, opened 2019-02-02
Use slices of data read from disk

eolivelli:
With this patch data read from disk (data and indexes) won't be splitted into many small byte[], this should save resources and GC pressure.
The trade off is that we will be rataining hard refs to the whole byte array until we have at least one reference to a portion of an Record inside the data page

### DIFF
--- a/herddb-utils/src/main/java/herddb/utils/Bytes.java
+++ b/herddb-utils/src/main/java/herddb/utils/Bytes.java
@@ -240,8 +240,8 @@ public final class Bytes implements Comparable<Bytes>, SizeAwareObject {
             if (other.hashCode() != this.hashCode()) {
                 return false;
             }
-            return CompareBytesUtils.arraysEquals(buffer, offset, length,
-                    other.buffer, other.offset, other.length);
+            return CompareBytesUtils.arraysEquals(buffer, offset, offset + length,
+                    other.buffer, other.offset, other.offset + other.length);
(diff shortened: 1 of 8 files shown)


herddb-utils/src/main/java/herddb/utils/Bytes.java
```diff
@@ -240,8 +240,8 @@ public boolean equals(Object obj) {
             if (other.hashCode() != this.hashCode()) {
                 return false;
             }
-            return CompareBytesUtils.arraysEquals(buffer, offset, length,
-                    other.buffer, other.offset, other.length);
+            return CompareBytesUtils.arraysEquals(buffer, offset, offset + length,
```
eolivelli: This is a bug of the previous commit !

eolivelli:
@diegosalvi this is change is good to go.

I did not add a configuration flag because for the indexes patch I had to change the signature of an interface.
I this that this is our way, we will confirm with benchmarks that this change improves performances in cases of low memory (lots of swap in/swap out).

@hamadodene  please compare this version with 0.8.1 as soon as we merge this patch.
The change will mostly affect only the case in which live data cannot reside entirely in memory (when you have lots of *datapage_reads* in metrics)

eolivelli:
@diegosalvi coveralls.io is telling us that we have poor coverage of some blocks: they are not always touched by tests execution. 
We should add a bunch of tests in order to traverse such code paths at every execution of tests

PR #329, opened 2019-02-05
Do not keep stong references to the IO buffer for temporary pages

eolivelli:


eolivelli:
@diegosalvi  the patch is not solving current @hamadodene problems but it is a real fix to be committed now

PR #331, opened 2019-02-08
Speed up PK writes

eolivelli:
During benchmarks with low-memory we are observing that reads and writes of the BLink are the bottleneck (as expected).
We can work on saving memory copies in order to acheive better performances.

Summary of changes:
- during page writes allocate a byte[] buffer, write data and perform hash inside the buffer, then write to next layer
- this new temporary buffer is recyclable

Please note that the recyclable buffer pool is private to the FileDataStorageManager, because it will hold always buffers of the same size (bounded by the maximum page size).
If we provide a genertic RecyclableByteArrayOutputStream we will eventually end up with a lot of churn

### DIFF
--- a/herddb-core/src/main/java/herddb/file/FileDataStorageManager.java
+++ b/herddb-core/src/main/java/herddb/file/FileDataStorageManager.java
@@ -80,7 +80,9 @@ import herddb.utils.OpenFileUtils;
 import herddb.utils.SimpleBufferedOutputStream;
 import herddb.utils.SimpleByteArrayInputStream;
 import herddb.utils.SystemProperties;
+import herddb.utils.VisibleByteArrayOutputStream;
 import herddb.utils.XXHash64Utils;
+import io.netty.util.Recycler;
 
 /**
  * Data Storage on local filesystem
@@ -223,15 +225,15 @@ public class FileDataStorageManager extends DataStorageManager {
         return getTablespaceDirectory(tableSpace).resolve(indexname + ".index");
     }
 
-    private Path getPageFile(Path tableDirectory, Long pageId) {
+    private static Path getPageFile(Path tableDirectory, Long pageId) {
         return tableDirectory.resolve(pageId + FILEEXTENSION_PAGE);
     }
 
-    private Path getTableCheckPointsFile(Path tableDirectory, LogSequenceNumber sequenceNumber) {
+    private static Path getTableCheckPointsFile(Path tableDirectory, LogSequenceNumber sequenceNumber) {
         return tableDirectory.resolve(sequenceNumber.ledgerId + "." + sequenceNumber.offset + EXTENSION_TABLEORINDExCHECKPOINTINFOFILE);
     }
 
-    private boolean isTableOrIndexCheckpointsFile(Path path) {
+    private static boolean isTableOrIndexCheckpointsFile(Path path) {
         Path filename = path.getFileName();
         return filename != null && filename.toString().endsWith(EXTENSION_TABLEORINDExCHECKPOINTINFOFILE);
     }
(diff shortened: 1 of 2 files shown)


herddb-core/src/main/java/herddb/file/FileDataStorageManager.java
```diff
@@ -896,24 +897,18 @@ public void writePage(String tableSpace, String tableName, long pageId, Collecti
...
-            size = hash.size() - size;
-
-            // footer
-            dataOutput.writeLong(hash.hash());
-            dataOutput.flush();
-            file.sync();
-
-            return size;
-        }
+        VisibleByteArrayOutputStream oo = new VisibleByteArrayOutputStream(1024 * 1024);
```
eolivelli: write in memory

herddb-core/src/main/java/herddb/file/FileDataStorageManager.java
```diff
@@ -896,24 +897,18 @@ public void writePage(String tableSpace, String tableName, long pageId, Collecti
...
+        ExtendedDataOutputStream dataOutput = new ExtendedDataOutputStream(oo);
+        dataOutput.writeVLong(1); // version
+        dataOutput.writeVLong(0); // flags for future implementations
+        writer.write(dataOutput);
+        dataOutput.flush();
+        long hash = XXHash64Utils.hash(oo.getBuffer(), 0, oo.size());
```
eolivelli: perform hash on the byte[]

herddb-core/src/main/java/herddb/file/FileDataStorageManager.java
```diff
@@ -896,24 +897,18 @@ public void writePage(String tableSpace, String tableName, long pageId, Collecti
...
+        dataOutput.writeLong(hash);
+        dataOutput.flush();
+        stream.write(oo.getBuffer(), 0, oo.size());
```
eolivelli: write to disk

eolivelli:
@diegosalvi @hamadodene @aluccaroni with this patch I have a significant improvement:

before:
```
filedatastore_index_pagewrites{success="true",quantile="0.5"} 56.0
filedatastore_index_pagewrites{success="true",quantile="0.75"} 120.0
filedatastore_index_pagewrites{success="true",quantile="0.95"} 219.0
filedatastore_index_pagewrites{success="true",quantile="0.99"} 279.0
filedatastore_index_pagewrites{success="true",quantile="0.999"} 404.0
filedatastore_index_pagewrites{success="true",quantile="0.9999"} 404.0
filedatastore_index_pagewrites{success="true",quantile="1.0"} 417.0
```
after:
```
filedatastore_index_pagewrites{success="true",quantile="0.5"} 18.0
filedatastore_index_pagewrites{success="true",quantile="0.75"} 52.0
filedatastore_index_pagewrites{success="true",quantile="0.95"} 118.0
filedatastore_index_pagewrites{success="true",quantile="0.99"} 179.0
filedatastore_index_pagewrites{success="true",quantile="0.999"} 284.0
filedatastore_index_pagewrites{success="true",quantile="0.9999"} 286.0
filedatastore_index_pagewrites{success="true",quantile="1.0"} 543.0
```

There is a peak on 1.0 percentile, but other values are strictly better up to 0.9999

herddb-core/src/main/java/herddb/file/FileDataStorageManager.java
diegosalvi: This is a full page "double" buffer. We could try to recycle it.

eolivelli:
@diegosalvi  I have addressed your comment and I have applied the same change to the logic which writes the pages

PR #334, opened 2019-02-12
Remove shared references during checkpoint

diegosalvi:
Checkpoint procedure preserved shared Bytes references (putting them even in keyToPage and in new pages during dirty page rebuild)


herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -797,10 +797,10 @@ private long flushNewPage(DataPage page, Map<Bytes, Record> spareData) {
         boolean add = true;
         final Iterator<Record> records = spareData.values().iterator();
         while (add && records.hasNext()) {
-            Record record = records.next();
+            Record record = records.next().nonShared();
             add = page.put(record);
```
eolivelli: This is going to allocate a new Record object

diegosalvi: Record is a wrapper around to Bytes. This will allocate a new record if bytes are shared.
Otherwise we sould set non final key and value inside record and replace them.

PR #338, opened 2019-02-18
Checkpoint: Keep rebuilt pages in memory only if needed

diegosalvi:
During checkpoints dirty and small rebuilt pages are added to page replacement policy and kept loaded in memory it shouldn't be se: records from really old unloaded pages dirtied for a record removal or small unloaded pages should simply remain unloaded (checkpoint should't change loaded page set, just reorganize it)

### DIFF
--- a/herddb-core/src/main/java/herddb/core/TableManager.java
+++ b/herddb-core/src/main/java/herddb/core/TableManager.java
@@ -2173,7 +2173,7 @@ public final class TableManager implements AbstractTableManager, Page.Owner {
                     /* Flush the page if it would exceed max page size */
                     final long recordSize = DataPage.estimateEntrySize(record);
                     if (bufferPageSize + recordSize > maxLogicalPageSize) {
-                        /* Keep in memory? Actually yes */
+                        /* Do no keep in memory old untouched records */
                         createImmutablePage(buffer, bufferPageSize, keepFlushedPageInMemory);
 
                         /* Reset next rebuilt page status */
@@ -2225,7 +2225,14 @@ public final class TableManager implements AbstractTableManager, Page.Owner {
                 }
             }
 
-            /* Flush remaining records */
+            /*
+             * Flush remaining records.
+             *
+             * To keep or not flushed page in memory is a "best guess" here: we don't known if records that
+             * needed to be kept in memory were already be flushed during newPage filling (see
+             * flushNewPageForCheckpoint). So we still use keepFlushedPageInMemory (possibily true) even if
+             * remaining records came from an old unused page.
+             */
             if (!buffer.isEmpty()) {
                 createImmutablePage(buffer, bufferPageSize, /* best guess */ keepFlushedPageInMemory);
                 flushedRecords += buffer.size();
--- a/herddb-core/src/main/java/herddb/core/TableManager.java
+++ b/herddb-core/src/main/java/herddb/core/TableManager.java
@@ -553,7 +553,7 @@ public final class TableManager implements AbstractTableManager, Page.Owner {
      * during checkpoint: <b>during "stop-the-world" procedures!</b>
      * </p>
      */
-    private long createImmutablePage(Map<Bytes, Record> newPage, long newPageSize) throws DataStorageManagerException {
+    private long createImmutablePage(Map<Bytes, Record> newPage, long newPageSize, boolean keepInMemory) throws DataStorageManagerException {
         final Long pageId = nextPageId++;
         final DataPage dataPage = buildImmutableDataPage(pageId, newPage, newPageSize);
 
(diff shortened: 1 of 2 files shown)


herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -2054,20 +2063,33 @@ private TableCheckpoint checkpoint(double dirtyThreshold, double fillThreshold,
...
             /* Rebuild dirty pages with only records to be kept */
             for (WeightedPage weighted : flushingDirtyPages) {
 
                 /* Page flushed */
                 flushedPages.add(weighted.pageId);
                 ++flushedDirtyPages;
 
-                final DataPage dataPage = pages.get(weighted.pageId);
+                final boolean currentPageWasInMemory;
+                final DataPage dataPage = pages.remove(weighted.pageId);
```
eolivelli: is this 'remove' needed ?

I am thinking about the case in which the checkpoint is running and we have a query (currently this is not possible but in the future it will be), if you remove the pages from 'pages' the page won't be more usable.
Shouldn't we remove the page only after appending the new pages ?

herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -2080,13 +2102,21 @@ private TableCheckpoint checkpoint(double dirtyThreshold, double fillThreshold,
...
+                        keepFlushedPageInMemory = false;
+
                         flushedRecords += buffer.size();
                         bufferPageSize = 0;
                         /* Do not clean old buffer! It will used in generated pages to avoid too many copies! */
                         buffer = new HashMap<>(buffer.size());
                     }
 
+                    /* Current rebuilt page will be kept in memory if current page was in memory */
+                    keepFlushedPageInMemory |= currentPageWasInMemory;
```
eolivelli: why are we keeping the last page ? 
are we going to modify it and not creating a new "current page" after the checkpoint ?

herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -2120,26 +2150,44 @@ private TableCheckpoint checkpoint(double dirtyThreshold, double fillThreshold,
...
+                        pageReplacementPolicy.remove(dataPage);
+                    }
                 }
 
                 for (Record record : records) {
                     /* Flush the page if it would exceed max page size */
                     final long recordSize = DataPage.estimateEntrySize(record);
                     if (bufferPageSize + recordSize > maxLogicalPageSize) {
-                        createImmutablePage(buffer, bufferPageSize);
+                        /* Keep in memory? Actually yes */
```
eolivelli: this comment is not clear.
if you are sure that keepFlushedPageInMemory is always true to be it is better to:
- add an assertion
- use the constant 'true' /* keepFlushedPageInMemory */ in the line 2177 below

herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -2172,15 +2220,14 @@ private TableCheckpoint checkpoint(double dirtyThreshold, double fillThreshold,
...
-//                    dataPage.makeImmutable();
                     ++flushedNewPages;
                     flushedRecords += dataPage.size();
                 }
             }
 
             /* Flush remaining records */
             if (!buffer.isEmpty()) {
-                createImmutablePage(buffer, bufferPageSize);
+                createImmutablePage(buffer, bufferPageSize, /* best guess */ keepFlushedPageInMemory);
```
eolivelli: can you explain better "best guess" ?
tomorrow we won't remebber about this patch, it is better to add a comment here

herddb-core/src/main/java/herddb/core/TableManager.java
diegosalvi: remaining "spare" data will be flushed into newpages if there is enough space. If there is still too "spare" date it will be flushed on a new immutable page after new pages flush. (This behaviour is unchanged, what chaged here is that if a buffer don't contains data from any loaded page the page created from it shouldn't be kept in memory)

herddb-core/src/main/java/herddb/core/TableManager.java
diegosalvi: Is a dev leftover... It should be /* Do no keep in memory old untouched records */ as line 2105
I'll change it

herddb-core/src/main/java/herddb/core/TableManager.java
diegosalvi: I'll reword better. It's just a fast note written 1 min before commit :(

This is a "best guess" because we don't known if records that needed to be kept in memory were already be flushed during newPage filling (line 2222). So we still use keepFlushedPageInMemory (possibily true) even if remaining records came from an old unused page.

herddb-core/src/main/java/herddb/core/TableManager.java
diegosalvi: Log fixed

herddb-core/src/main/java/herddb/core/TableManager.java
diegosalvi: Added a more explaining comment

diegosalvi:
CI failed due to a timeout during submission to coveralls (tests were ok)
`[ERROR] Submission failed in 927ms while processing data`

eolivelli:
Okay.
Go ahead and merge :)

PR #340, opened 2019-02-19
Fix/remove checkpoint double pk lookup closes #333

### LINKED ISSUE #333: Checkpoint: reduce PK lookup
During checkpoint dirty page cleanup there is a double PK lookup in primary index:
- first one to discover if the row is "good":
`if (currentPageId == null || !weighted.pageId.equals(currentPageId)) {`
- the second one in page creation (`createImmutablePage`) to overwrite the page in the primary index:
`keyToPage.put(key, pageId);`

We could rewrite the procedure to avoid double PK lookup adding a method putIf(key,pageID,expectedPageID). It should permit to check if the record must be moved away and insert the new value at the same time

diegosalvi:
This pull requext resolve #333

A new method on KeyToPageIndex has been added to update index values ONLY if current value is as expected parameter (like atomic compare and set). Such method is used during dirty page rebuild to avoid a double lookup (1st to known is current record is "old" and a 2nd one to update the page reference).

To be able to udpate page reference, page id "generation" has been moved in checkpoint code and there maintained.

**This patch must be merged after #338 (is build on top of it)**

### DIFF
--- a/herddb-core/src/main/java/herddb/core/DataPage.java
+++ b/herddb-core/src/main/java/herddb/core/DataPage.java
@@ -83,6 +83,32 @@ public class DataPage extends Page<TableManager> {
         pageLock = immutable ? null : new ReentrantReadWriteLock(false);
     }
 
+    /**
+     * Convert a {@link DataPage} to immutable.
+     * <p>
+     * Conversion can be done only after the page has been set as not writable.
+     * </p>
+     * <p>
+     * Checks on immutable pages are faster (they not have to check volatile writable flag and lock it
+     * before checking).
+     * </p>
+     *
+     * @return immutable data page version
+     */
+    DataPage toImmutable() {
+
+        if (immutable) {
+            throw new IllegalStateException("page " + pageId + " already is immutable!");
+        }
+
+        if (writable) {
+            throw new IllegalStateException("page " + pageId + " cannot be converted to immutable because still writable!");
+        }
+
+        return new DataPage(owner, pageId, maxSize, usedMemory.get(), data, true);
+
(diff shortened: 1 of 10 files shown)


diegosalvi:
Rebased on master

diegosalvi:
I'll mobile check. I didn't have Ant problems on my pc

Il Gio 21 Feb 2019, 16:10 Enrico Olivelli <notifications@github.com> ha
scritto:

diegosalvi:
Double not mobile...

Il Gio 21 Feb 2019, 16:11 Lothruin Mirwen <lothruin.mirwen@gmail.com> ha
scritto:

diegosalvi:
I cannot replicate the problem but I presume that is related with scan procedure. Actually checkpoint execute presuming a "stop the world" procedure (for both read and writes) but is isn't like that: scan do not take the checkpointLock so concurrent checkpoints can be done. Fetched records and pages can be moved away during scan (and record fetch) generating the problem.

We could force pause of scans too or abort this patch (to do just one PK access we need to checkAndSet the PK so publishing on PK records from unflushed/uncreated pages)

diegosalvi:
Pushed a modification on record memory "publish"

Still a WIP

diegosalvi:
Added a concurrent test for "putIf" in key to page.
Fixed an error on MemoryDataStorageManager refusing index pages to be rewritten (actually BLink rewrite pages when it known they arent checkpoint pages)

diegosalvi:
Commented code removed, my bad :(

The stacktrace can be kept to known why the test exists in the firts place.

PR #345, opened 2019-03-01
Support non null constraints

amitvc:
Added support for non null types for the following data types *long, int, string*. Also added several unit tests to validate that changes work. I have added a new method in Table.java to validate primary key based on the data type. If I make the data type check stricter to only not_null members then a lot of code needs to be changed and UT's need to modified. I think we should handle that situation in different PR. Some UT's seem to be broken.


amitvc:
@eolivelli
*Note*
Unit test in AlterTableSQLTest.modifyColumn, renameTable fail due to assertion errors
Sample below
	at herddb.core.DBManager.executePlan(DBManager.java:672)
	at herddb.core.TestUtils.execute(TestUtils.java:66)
	at herddb.core.TestUtils.execute(TestUtils.java:61)
	at herddb.core.AlterTableSQLTest.modifyColumn(AlterTableSQLTest.java:165)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
Caused by: java.lang.AssertionError
	at herddb.utils.Bytes.to_long(Bytes.java:168)
	at herddb.core.TableManager.applyInsert(TableManager.java:1694)
	at herddb.core.TableManager.apply(TableManager.java:1388)
	at herddb.core.TableManager.lambda$executeInsertAsync$9(TableManager.java:944)
	at java.util.concurrent.CompletableFuture.uniApply(CompletableFuture.java:602)
	at

amitvc:
I am trying to figure out why they are failing. But I will look into it tomorrow. Any advice?

herddb-core/src/main/java/herddb/codec/RecordSerializer.java
```diff
@@ -282,12 +304,13 @@ public static void skipTypeAndValue(ByteArrayCursor dii) throws IOException {
 
     public static byte[] serialize(Object v, int type) {
         if (v == null) {
-            return null;
+            throw new IllegalArgumentException("Object to be serialized cannot be null");
```
eolivelli: we cannot change this behevious
and NULL is a valid value to be serialized

herddb-core/src/main/java/herddb/codec/RecordSerializer.java
```diff
@@ -282,12 +304,13 @@ public static void skipTypeAndValue(ByteArrayCursor dii) throws IOException {
...
         }
         switch (type) {
             case ColumnTypes.BYTEARRAY:
                 return (byte[]) v;
             case ColumnTypes.INTEGER:
+            case ColumnTypes.NOTNULL_INTEGER:
```
eolivelli: here you cannot serialize a NULL, for a NOTNULL_INTEGER, you have to duplicate all cases

herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -389,6 +389,18 @@ public Table getTable() {
                     return table;
                 }
             };
+        } else if (table.getColumn(table.primaryKey[0]).type == ColumnTypes.NOTNULL_INTEGER) {
```
eolivelli: can you merge this branch with the ColumnTypes.INTEGER one ?

herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -401,6 +413,18 @@ public Table getTable() {
                     return table;
                 }
             };
+        } else if (table.getColumn(table.primaryKey[0]).type == ColumnTypes.NOTNULL_LONG) {
```
eolivelli: can you merge this branch with the ColumnTypes.LONG one ?

herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -1667,10 +1691,10 @@ private void applyInsert(Bytes key, Bytes value, boolean onTransaction) throws D
 
         if (table.auto_increment) {
             // the next auto_increment value MUST be greater than every other explict value
-            long pk_logical_value;
-            if (table.getColumn(table.primaryKey[0]).type == ColumnTypes.INTEGER) {
+            long pk_logical_value = key.to_long();
```
eolivelli: in case of an int you have to use key.to_int(), this will result in some bad runtime error (maybe the error in your test)
because you can convert only a 8-bytes Bytes into a long, not a 4-bytes one

herddb-core/src/main/java/herddb/core/system/SyscolumnsTableManager.java
```diff
@@ -39,8 +39,8 @@
     private final static Table TABLE = Table
             .builder()
             .name("syscolumns")
-            .column("table_name", ColumnTypes.STRING)
-            .column("column_name", ColumnTypes.STRING)
+            .column("table_name", ColumnTypes.NOTNULL_STRING)
```
eolivelli: This MAY break compatibily with older clients (which are not aware of NOT NULL columns).
I would defer this change on a new PR

herddb-core/src/main/java/herddb/core/system/SysconfigTableManager.java
```diff
@@ -37,7 +37,7 @@
     private final static Table TABLE = Table
         .builder()
         .name("sysconfig")
-        .column("name", ColumnTypes.STRING)
+        .column("name", ColumnTypes.NOTNULL_STRING)
```
eolivelli: same here

herddb-core/src/main/java/herddb/core/system/SysindexcolumnsTableManager.java
```diff
@@ -45,11 +45,11 @@
             .builder()
             .name("sysindexcolumns")
             .column("tablespace", ColumnTypes.STRING)
-            .column("table_name", ColumnTypes.STRING)
-            .column("index_name", ColumnTypes.STRING)
+            .column("table_name", ColumnTypes.NOTNULL_STRING)
```
eolivelli: same here and in all of the other systables

herddb-core/src/main/java/herddb/model/ColumnTypes.java
```diff
@@ -34,8 +34,11 @@
     public static final int NULL = 5;
     public static final int DOUBLE = 6;
     public static final int BOOLEAN = 7;
+    public static final int NOTNULL_STRING = 9;
+    public static final int NOTNULL_INTEGER = 10;
+    public static final int NOTNULL_LONG = 11;
 
-    public static final int ANYTYPE = 10;
+    public static final int ANYTYPE = 12;
```
eolivelli: Do not change values of constants, java compiler eagerly copies values to .class files, we will break compatibility.

herddb-core/src/main/java/herddb/model/Table.java
```diff
@@ -440,11 +440,8 @@ public Table build() {
                 if (pk == null) {
                     throw new IllegalArgumentException("column " + pkColumn + " is not defined in table");
                 }
-                if (pk.type != ColumnTypes.STRING
-                        && pk.type != ColumnTypes.LONG
-                        && pk.type != ColumnTypes.INTEGER
-                        && pk.type != ColumnTypes.TIMESTAMP) {
-                    throw new IllegalArgumentException("primary key " + pkColumn + " must be a string or long or integer or timestamp");
+                if (!validatePrimaryKeyDataType(pk.type)) {
+                    throw new IllegalArgumentException("primary key " +pkColumn+ " must be a non null string or long or integer or timestamp");
```
eolivelli: you can leave the old message, it is implicit that PKs are not null

herddb-core/src/main/java/herddb/model/Table.java
```diff
@@ -455,6 +452,21 @@ public Table build() {
                     tablespace, auto_increment, maxSerialPosition);
         }
 
+        private static boolean validatePrimaryKeyDataType(int type) {
+            switch(type) {
+                case ColumnTypes.NOTNULL_INTEGER:
+                case ColumnTypes.INTEGER:
```
eolivelli: please add a comment to explain that we are allowing nullable data types in order to keep compatibility with pre-0.9 databases and clients

herddb-core/src/test/java/herddb/codec/RecordSerializerTest.java
```diff
@@ -47,7 +49,7 @@ public RecordSerializerTest() {
     public void testToBean() {
         Table table = Table.builder()
             .name("t1")
-            .column("pk", ColumnTypes.STRING)
+            .column("pk", ColumnTypes.NOTNULL_STRING)
```
eolivelli: do not change this test, add a new one if you think it has some value

herddb-core/src/test/java/herddb/core/CreateTableTest.java
```diff
@@ -53,7 +53,7 @@ public void createTable1() throws Exception {
                     .builder()
                     .tablespace("tblspace1")
                     .name("t1")
-                    .column("id", ColumnTypes.STRING)
+                    .column("id", ColumnTypes.NOTNULL_STRING)
```
eolivelli: same as above

herddb-core/src/test/java/herddb/core/CreateTableTest.java
```diff
@@ -63,4 +63,50 @@ public void createTable1() throws Exception {
         }
 
     }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testWithNullPrimaryKey() throws Exception {
```
eolivelli: Not sure about the value of this test.
For backward compatibility we have to allow PKs which are defined with NULLABLE column types

eolivelli: or is this test about using a DOUBLE column ?

herddb-core/src/test/java/herddb/core/CreateTableTest.java
```diff
@@ -63,4 +63,50 @@ public void createTable1() throws Exception {
...
+                    .primaryKey("id")
+                    .build();
+
+            CreateTableStatement st2 = new CreateTableStatement(table);
+            manager.executeStatement(st2, StatementEvaluationContext.DEFAULT_EVALUATION_CONTEXT(), TransactionContext.NO_TRANSACTION);
+        }
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testWithNullPrimaryKey2() throws Exception {
```
eolivelli: same as above

herddb-core/src/main/java/herddb/codec/RecordSerializer.java
amitvc: Done. I reverted this.

herddb-core/src/main/java/herddb/codec/RecordSerializer.java
amitvc: Done.

herddb-core/src/main/java/herddb/core/TableManager.java
amitvc: Done.

herddb-core/src/main/java/herddb/core/TableManager.java
amitvc: 👍

herddb-core/src/main/java/herddb/core/TableManager.java
amitvc: Yes. Good catch. I will fix this.

herddb-core/src/main/java/herddb/core/system/SyscolumnsTableManager.java
amitvc: I reverted both *table_name* and *column_name* changes for this file.

herddb-core/src/main/java/herddb/core/system/SysconfigTableManager.java
amitvc: 👍

herddb-core/src/main/java/herddb/model/Table.java
amitvc: 👍

herddb-core/src/main/java/herddb/model/ColumnTypes.java
amitvc: Agreed. I changed it back and moved the new types after value 10.

herddb-core/src/test/java/herddb/codec/RecordSerializerTest.java
amitvc: 👍

herddb-core/src/test/java/herddb/core/CreateTableTest.java
amitvc: done.

herddb-core/src/test/java/herddb/core/CreateTableTest.java
amitvc: This test is to ensure we throw exception when primary key is chosen as Double.

herddb-core/src/test/java/herddb/core/CreateTableTest.java
amitvc: This test is to ensure we throw exception when primary key is chosen as ByteArray

amitvc:
@eolivelli - Addressed code review comments. Now all tests pass.

herddb-core/src/main/java/herddb/core/system/SysclientsTableManager.java
```diff
@@ -38,7 +38,7 @@
     private final static Table TABLE = Table
             .builder()
             .name("sysclients")
-            .column("id", ColumnTypes.STRING)
+            .column("id", ColumnTypes.NOTNULL_STRING)
```
eolivelli: you have to revert this change as well

herddb-core/src/main/java/herddb/core/system/SyslogstatusManager.java
```diff
@@ -39,8 +39,8 @@
     private final static Table TABLE = Table
             .builder()
             .name("syslogstatus")
-            .column("tablespace_uuid", ColumnTypes.STRING)
-            .column("nodeid", ColumnTypes.STRING)
+            .column("tablespace_uuid", ColumnTypes.NOTNULL_STRING)
```
eolivelli: you have to revert this change as well and the line below

herddb-core/src/main/java/herddb/core/system/SysnodesTableManager.java
```diff
@@ -41,7 +41,7 @@
     private final static Table TABLE = Table
             .builder()
             .name("sysnodes")
-            .column("nodeid", ColumnTypes.STRING)
+            .column("nodeid", ColumnTypes.NOTNULL_STRING)
```
eolivelli: you have to revert this change as well

herddb-core/src/main/java/herddb/core/system/SysstatementsTableManager.java
```diff
@@ -39,7 +39,7 @@
     private final static Table TABLE = Table
             .builder()
             .name("sysstatements")
-            .column("id", ColumnTypes.LONG)
+            .column("id", ColumnTypes.NOTNULL_LONG)
```
eolivelli: you have to revert this change as well

herddb-core/src/main/java/herddb/core/system/SystablesTableManager.java
```diff
@@ -37,8 +37,8 @@
     private final static Table TABLE = Table
             .builder()
             .name("systables")
-            .column("tablespace", ColumnTypes.STRING)
-            .column("table_name", ColumnTypes.STRING)
+            .column("tablespace", ColumnTypes.NOTNULL_STRING)
```
eolivelli: you have to revert this change as well and the line below

herddb-core/src/main/java/herddb/core/system/SystablespacereplicastateTableManager.java
```diff
@@ -43,8 +43,8 @@
         .builder()
         .name("systablespacereplicastate")
         .column("tablespace_name", ColumnTypes.STRING)
-        .column("uuid", ColumnTypes.STRING)
-        .column("nodeid", ColumnTypes.STRING)
+        .column("uuid", ColumnTypes.NOTNULL_STRING)
```
eolivelli: you have to revert this change as well and the line below

herddb-core/src/main/java/herddb/core/system/SystransactionsTableManager.java
```diff
@@ -39,7 +39,7 @@
         .builder()
         .name("systransactions")
         .column("tablespace", ColumnTypes.STRING)
-        .column("txid", ColumnTypes.LONG)
+        .column("txid", ColumnTypes.NOTNULL_LONG)
```
eolivelli: you have to revert this change as well

herddb-core/src/test/java/herddb/codec/RecordSerializerTest.java
```diff
@@ -79,4 +80,52 @@ private static void testTimestamp(String testCase, String timezone, long expecte
         assertEquals("failed for " + testCase + " delta is " + delta + " h, result is " + formattedResult, expectedResult, result.getTime());
     }
 
+    @Test
+    public void testSerializeWithNullAndNonNullTypes() {
```
eolivelli: we are missing tests about serializing a NOT_NULL_INTEGER (and STRING...all of the types) and throw an IllegalArgumentException.

This is the key of this implementation, this way we are sure that for a NON_NULL_XXX you won't ever have a NULL.

herddb-core/src/test/java/herddb/codec/RecordSerializerTest.java
```diff
@@ -79,4 +80,52 @@ private static void testTimestamp(String testCase, String timezone, long expecte
...
+    }
+
+    @Test
+    public void testSerializeThrowsExceptionOnNullObject() {
+        assertNull(RecordSerializer.serialize(null, ColumnTypes.STRING));
+    }
+
+    @Test
+    public void testDeserializeWithNullAndNonNullTypes() {
+        byte iValueAsByteArray[] = ByteBuffer.allocate(4).putInt(1000).array();
```
eolivelli: you have to use h.u.Bytes.from_int, other wise you won't handle endianness consistently

don't use ByteBuffer, but always h.u.Bytes, even in the lines below

herddb-core/src/test/java/herddb/core/CreateTableTest.java
eolivelli: please rename accordingly

herddb-core/src/test/java/herddb/core/CreateTableTest.java
eolivelli: please rename

herddb-core/src/test/java/herddb/core/CreateTableTest.java
amitvc: Done.

herddb-core/src/test/java/herddb/core/CreateTableTest.java
amitvc: Done.

herddb-core/src/main/java/herddb/core/system/SysclientsTableManager.java
amitvc: 👍

herddb-core/src/main/java/herddb/core/system/SyslogstatusManager.java
amitvc: Done.

herddb-core/src/main/java/herddb/core/system/SysnodesTableManager.java
amitvc: Done

herddb-core/src/main/java/herddb/core/system/SystablesTableManager.java
amitvc: Done.

herddb-core/src/main/java/herddb/core/system/SystransactionsTableManager.java
amitvc: Done.

herddb-core/src/main/java/herddb/core/system/SysstatementsTableManager.java
amitvc: Done.

herddb-core/src/test/java/herddb/codec/RecordSerializerTest.java
amitvc: I think there might be a problem. Currently we return null if the object is null from the serialize method. 
public static byte[] serialize(Object v, int type) {
        if (v == null) {
            return null;
        }

herddb-core/src/test/java/herddb/core/CreateTableTest.java
```diff
@@ -105,8 +106,67 @@ public void testWithNullPrimaryKey2() throws Exception {
...
+                    .primaryKey("id")
+                    .build();
+
             CreateTableStatement st2 = new CreateTableStatement(table);
             manager.executeStatement(st2, StatementEvaluationContext.DEFAULT_EVALUATION_CONTEXT(), TransactionContext.NO_TRANSACTION);
+            TranslatedQuery translated = manager.getPlanner().translate("tblspace1", "INSERT INTO t1 (id,name) values(?,?)", Arrays.asList(null, 1234), true, true, false, -1);
+            manager.executePlan(translated.plan, translated.context, TransactionContext.NO_TRANSACTION);
         }
     }
+
```
amitvc: The reason for this UT to be commented out is -
1. I have some observations here. The null check on column only works if they are part of the primary key. How can we enforce the same if non primary key columns have data types which are not null? 
2. I don't see an easy way to add a test for update query with null value for the primary key. It just ignore it and keeps the original record.

amitvc:
@eolivelli I have added some comments and addressed the code review comments and made necessary changes.

herddb-core/src/test/java/herddb/core/CreateTableTest.java
eolivelli: Actually you cannot update the PK (it is better to file an issue).
It would be like a DELETE + INSERT, but it is a case not actually covered.

> The null check on column only works if they are part of the primary key
I am expecting that with your patch any NULL in a NOT-NULL column must fail the serialization of the record, and this will lead to an error to the client.

If this is not happening check with a break point here 
https://github.com/diennea/herddb/blob/6393743bafb800c7e86fdd86b52019d1903a0bb2/herddb-core/src/main/java/herddb/core/TableManager.java#L964

We are catching any "StatementExecutionException" and rethrowing it as a validation error, maybe you have to throw a StatementExecutionException during the serialization of the record

eolivelli: This is where you have to put the new validation:
https://github.com/diennea/herddb/blob/aa7e742d2e9664513745630f94198ca2e788bca0/herddb-core/src/main/java/herddb/codec/RecordSerializer.java#L306

```
if (v == null) {
    if (ColumnTypes.isNotNullDataType(type) {
           throw new StatementExecutionException("Cannot serialize null of a "+typeToString(type)+" column type);
} else {
    return null;
} 
}
```

the same for serializeTo() and for validate() and for serializeValue()

amitvc: Good suggestion.

amitvc:
@eolivelli -- I see couple of things.
I am creating a table with the following schema - 
Table table = Table
                    .builder()
                    .tablespace("tblspace1")
                    .name("t1")
                    .column("id", ColumnTypes.STRING)
                    .column("name", ColumnTypes.NOTNULL_STRING)
                    .primaryKey("id")
                    .build();
What I notice is while debugging **CalcitePlanner.convertToHerdType** the input to this function RelDataType is of type BasicSqlDataType. For the primary key  *id* I see the relType value to be VARCHAR while for *name* it is ANY. I dug more into this. It seems there SqlTypeName.java inside calcite library there is no not null data type. So it defaults to ANY. The reason I was looking into it is I was trying to throw an exception when we determine that on update/insert query if we have a column which is non-null but the value we are updating to or insert is null. Since CalcitePlanner tells us the type is ANY and we are trying to map the type to herddb type it never errors out. I think I am missing something here. I thought the SQLRecordPredicate cast method would have been perfect to bubble up the SQLExecutionException [null value in non-null column] but that did not work out since the types dont match. 
public static Object cast(Object value, int type) {
        if (value == null) {
            if (ColumnTypes.isNotNullDataType(type)) {
                throw new StatementExecutionException("Cannot serialize null of a "+ColumnTypes.typeToString(type)+" column type");
            }
            return null;
        }
      

Any thoughts on this ?

eolivelli:
@amitvc you have to work in CalcitePlanner#convertTypeNotNull and map all of the new datatypes, actually you miss a change to that method.

https://github.com/diennea/herddb/blob/6393743bafb800c7e86fdd86b52019d1903a0bb2/herddb-core/src/main/java/herddb/sql/CalcitePlanner.java#L1286

I don't expect DML statements to fail due to "cast" but due to "serialize" for trivial queries (but maybe I am wrong)

eolivelli:
Let CalcitePlanner.convertToHerdType return only "NULLABLE" datatypes, I don't think this is a real problem.

For Calcite what is really important is this line:
https://github.com/diennea/herddb/blob/6393743bafb800c7e86fdd86b52019d1903a0bb2/herddb-core/src/main/java/herddb/sql/CalcitePlanner.java#L1230

amitvc:
@eolivelli - Added unit tests and made suggested changes. Had trouble with Update query which was not releasing locks correctly when StateExecutionException was being thrown from record serializer. Added some unit test around that too.

herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -994,40 +996,44 @@ public default void endNewRecordsInTransactionBlock() {
...
+                    LogEntry entry = LogEntryFactory.update(table, actual.key, Bytes.from_array(newValue), transaction);
+                    CommitLogResult pos = log.log(entry, entry.transactionId <= 0);
+                    writes.add(pos.logSequenceNumber.thenApply(lsn -> new PendingLogEntryWork (entry, pos, lockHandle)));
+                    lastKey.value = actual.key;
+                    lastValue.value = newValue;
+                    updateCount.incrementAndGet();
+                }
+            }, transaction, true, true);
+        } catch (StatementExecutionException validationError) {
+            return FutureUtils.exception(validationError);
```
eolivelli: This is not the right point

That "validationError" should be handled in the same "catch" al line 1101 

your new error is throws from
`byte[] newValue = function.computeNewValue(actual, context, tableContext);`

we are already catching validation errors for 
`RecordSerializer.validatePrimaryKey`

you should enclose
byte[] newValue = function.computeNewValue(actual, context, tableContext);

in the same try/catch.
Your "serialization" error should be an IllegalArgumentException, this way we will handle consistently your new impl (not null columns) and the fact that indexed columns cannot be null (this is only because our BRIN index impl does support null keys, but this is another story).

amitvc: @eolivelli done with the changes. All tests pass.

amitvc:
@eolivelli thank you. Will start work on the next ticket tonight.

PR #354, opened 2019-03-12
Fix/checkpoint isolate methods

diegosalvi:
This PR starts from Enrico's fix/fix-checkpoint branch keeping all his improvements and isolating the main handling block of checkpoint into a separate method.

In addition:
* now dirty small pages handling is separated from dirty pages handling (thus a "small" and dirty page can be cleaned up without waiting forever that any other page has been cleaned up)
* added more DataPage isolation with a new metod to force used memory (without direct access to internal atomic long, cosmetic)
* fully handled "buildingPage" data size (who add or remove data from building page now handles page memory count too thus at the end of method execution buildPage data must match buildingPage usedMemory count with no other external count)

### DIFF
--- a/herddb-core/src/main/java/herddb/core/TableManager.java
+++ b/herddb-core/src/main/java/herddb/core/TableManager.java
@@ -713,10 +713,10 @@ public final class TableManager implements AbstractTableManager, Page.Owner {
      * </p>
      *
      * @param page new page to flush
-     * @param stealingDataPage old spare data to fit in the new page if possible
+     * @param spareDataPage old spare data to fit in the new page if possible
      */
-    private void flushNewPageForCheckpoint(DataPage page, DataPage stealingDataPage) {
-        final boolean flushed = flushNewPage(page, stealingDataPage);
+    private void flushNewPageForCheckpoint(DataPage page, DataPage spareDataPage) {
+        final boolean flushed = flushNewPage(page, spareDataPage);
 
         if (!flushed) {
             throw new IllegalStateException("Concurrently flushed a new page during checkpoint! Page " + page);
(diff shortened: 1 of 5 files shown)


herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -2043,6 +2061,146 @@ private static final long sumOverflowWise(long a, long b) {
...
+            this.buildingPage = buildingPage;
+            this.keepFlushedPageInMemory = keepFlushedPageInMemory;
+            this.flushedPages = flushedPages;
+            this.flushedRecords = flushedRecords;
+        }
+
+    }
+
+    /**
+     *
```
eolivelli: doc ?

herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -2043,6 +2061,146 @@ private static final long sumOverflowWise(long a, long b) {
...
+    private CleanAndCompactResult cleanAndCompactPages(List<WeightedPage> workingPages, DataPage buildingPage,
+                                                       boolean keepFlushedPageInMemory, long processLimitInstant) {
+
+        long buildingPageSize = buildingPage.getUsedMemory();
+
+        List<Long> flushedPages = new ArrayList<>();
+        long flushedRecords = 0;
+
+        /* Rebuild dirty pages with only records to be kept */
+        for (WeightedPage weighted : workingPages) {
```
eolivelli: what about renaming the variable "weighted"  to "page" or "pageToRebuild" or something more meaning full ?

herddb-core/src/main/java/herddb/core/TableManager.java
diegosalvi: added method doc

herddb-core/src/main/java/herddb/core/TableManager.java
diegosalvi: renamed just to page

diegosalvi:
Found the problem! It was introduced with building page lock removal during checkpoint.
We canno add the record to page before knowing if it can be added or we can overwrite the right record with an old version (and then remove it too)

PR #355, opened 2019-03-12
Fix/wrong checkpoint recovery

diegosalvi:
Changes:
* added a check to evaluate if inserted record is really new (it breaks some test, especially RestartTest.recoverTableCreatedInTransaction)
* added a check to skip transaction apply if table booted with newer data (log in recovery, failure otherwise)
* added some minor cosmetic toString method to simplify debugging

### DIFF
--- a/herddb-core/src/main/java/herddb/core/TableManager.java
+++ b/herddb-core/src/main/java/herddb/core/TableManager.java
@@ -1324,6 +1324,20 @@ public final class TableManager implements AbstractTableManager, Page.Owner {
             forceFlushTableData = true;
         }
 
+        if (!transaction.lastSequenceNumber.after(bootSequenceNumber)) {
+            if (recovery) {
+                LOGGER.log(Level.FINER,
+                        "ignoring transaction {0} commit on recovery, {1}.{2} data is newer: transaction {3}, table {4}",
+                        new Object[] { transaction.transactionId, table.tablespace, table.name,
+                                transaction.lastSequenceNumber, bootSequenceNumber });
+                return;
+            } else {
+                throw new DataStorageManagerException("corrupted commit log " + table.tablespace + "." + table.name
+                        + " data is newer than transaction " + transaction.transactionId + " transaction "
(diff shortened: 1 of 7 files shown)


herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -1324,6 +1342,20 @@ public void onTransactionCommit(Transaction transaction, boolean recovery) throw
             forceFlushTableData = true;
         }
 
+        if (!transaction.lastSequenceNumber.after(bootSequenceNumber)) {
```
eolivelli: Can we create a patch only for this single change? This way we will isolate the problem.
This branch is mixing too many fixed.
How does this sounds to you?

diegosalvi: Moved the patch on top of master, waiting for CI

herddb-core/src/main/java/herddb/model/Transaction.java
```diff
@@ -509,7 +507,7 @@ public boolean isOnTable(String name) {
                 || (newTables != null && newTables.containsKey(name));
     }
 
-    public void synch() throws LogNotAvailableException {
+    public void sync() throws LogNotAvailableException {
```
eolivelli: Is this one useful?

diegosalvi: Really for.. a test

eolivelli: Please add //visible for testing

diegosalvi: I'll add it... (I'll check if I can reduce visibility too)

diegosalvi:
There is a failing test due to an issue not directly related to this patch (just this patch trigger it on a test).
I'll push another PR for that

diegosalvi:
We should merge #357 before this to address PK memory tracking problems

PR #359, opened 2019-03-15
Add facility to handle non null column type

amitvc:

If your pull request is about ~20 lines of code you don't need to sign an iCLA if you are unsure
please ask on the developers list.

### DIFF
--- a/herddb-core/src/main/java/herddb/core/system/SyscolumnsTableManager.java
+++ b/herddb-core/src/main/java/herddb/core/system/SyscolumnsTableManager.java
@@ -70,7 +70,7 @@ public class SyscolumnsTableManager extends AbstractSystemTableManager {
                         "table_name", t.name,
                         "column_name", c.name,
                         "ordinal_position", pos++,
-                        "is_nullable", (pk || nonNullCType) ? DatabaseMetaData.columnNoNulls : DatabaseMetaData.columnNullable,
+                        "is_nullable", nonNullCType ? DatabaseMetaData.columnNoNulls : DatabaseMetaData.columnNullable,
                         "data_type", data_type,
                         "auto_increment", (pk && t.auto_increment)?1:0
                 ));
--- a/herddb-core/src/main/java/herddb/core/system/SyscolumnsTableManager.java
+++ b/herddb-core/src/main/java/herddb/core/system/SyscolumnsTableManager.java
@@ -19,6 +19,7 @@
  */
 package herddb.core.system;
 
+import java.sql.DatabaseMetaData;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -61,6 +62,7 @@ public class SyscolumnsTableManager extends AbstractSystemTableManager {
             int pos = 1;
             for (Column c : t.columns) {
                 boolean pk = t.isPrimaryKeyColumn(c.name);
+                boolean nonNullCType = pk ? true : ColumnTypes.isNotNullDataType(c.type);
                 String data_type = ColumnTypes.typeToString(c.type);
 
                 result.add(RecordSerializer.makeRecord(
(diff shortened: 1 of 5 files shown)


amitvc:
@eolivelli for the UT in jdbc module is it new class or I need to add more tests to existing classes of unit tests?

herddb-core/src/main/java/herddb/core/system/SyscolumnsTableManager.java
```diff
@@ -61,14 +62,15 @@ public SyscolumnsTableManager(TableSpaceManager parent) {
...
+                boolean nonNullCType = pk ? true : ColumnTypes.isNotNullDataType(c.type);
                 String data_type = ColumnTypes.typeToString(c.type);
 
                 result.add(RecordSerializer.makeRecord(
                         table,
                         "table_name", t.name,
                         "column_name", c.name,
                         "ordinal_position", pos++,
-                        "is_nullable", pk ? 0 : 1,
+                        "is_nullable", (pk || nonNullCType) ? DatabaseMetaData.columnNoNulls : DatabaseMetaData.columnNullable,
```
eolivelli: isn't "pl" reduntant here ?
if ("pk") nonNullCType is already true

eolivelli:
@amitvc good idea to split your work in smaller subtasks

eolivelli:
In the context of this patch you can add only a testcase in SystemTablesTest

JDBC clients won't be impacted directly, they will only "see" this new constraint, but it is already handled, as it was handled for pk columns.

herddb-core/src/main/java/herddb/core/system/SyscolumnsTableManager.java
amitvc: You are correct. I  removed it.

amitvc:
SystemTablesTest class is present in both core and jdbc. I am assuming adding tests in the one that is in the core project.

amitvc:
@eolivelli - I had to modify DDLSQLPlanner.java to add support for not null types when specified in query string. Also added UT to test my changes.

herddb-core/src/main/java/herddb/model/ColumnTypes.java
```diff
@@ -81,4 +81,17 @@ public static boolean isNotNullDataType(int type) {
...
+    public static int getNonNullTypeForPrimitiveType(int type) {
+        switch(type) {
+            case STRING:
+                return NOTNULL_STRING;
+            case INTEGER:
+                return NOTNULL_INTEGER;
+            case LONG:
+                return NOTNULL_LONG;
+            default:
+                return type;
```
eolivelli: not sure this is the best return value, what about am IllegalArgumentException ?
Other wide it is better to add a Javadoc

herddb-core/src/main/java/herddb/sql/DDLSQLPlanner.java
```diff
@@ -365,7 +364,14 @@ private Statement buildCreateTableStatement(String defaultTableSpace, CreateTabl
                     if (auto_increment && primaryKey.contains(cf.getColumnName())) {
                         tablebuilder.primaryKey(columnName, auto_increment);
                     }
+
+                    if(String.join("_",columnSpecs).equals("NOT_NULL")) {
+                        type = ColumnTypes.getNonNullTypeForPrimitiveType(type);
```
eolivelli: we should throw an error in case of a type which does not support NOT NULL
and we have to add an unit test around this case.

we can't let users try to create a constraint but this specs won't reflect the actual type of the column

herddb-core/src/main/java/herddb/model/ColumnTypes.java
amitvc: I modified the method to add more types that are supported in ColumnTypes.

herddb-core/src/main/java/herddb/sql/DDLSQLPlanner.java
amitvc: Good idea. Will add UT for this.

amitvc: @eolivelli Addressed changes and added UT

eolivelli:
I left one last comment

amitvc:
@eolivelli -  Yes the scan won't get executed. I removed it to make it more obvious what we are testing in the UT

amitvc:
@eolivelli thanks for the review.

PR #360, opened 2019-03-15
Force Build with a modern JDK

eolivelli:
Ensure we are compatible with Java 8


eolivelli:
@amitvc can your please review and test this branch on your dev env ?

With this change you will be forced to use at least JDK9+ to build HerdDB.

We are using JDK11 in our dev environments and on CI.
We have special code that leverages Hotspot intrinsics from Java10+ but we have to be compatible at run time with Java 8.

PR #364, opened 2019-03-25
Improved empty pages checkpoint handling. Closes #356

### LINKED ISSUE #356: Improve BLink empty pages handling
Actual BLink implementation never drop empty pages. They are removed from PageReplacementPolicy (only when requested from the page loading itself) and not written in metadata during checkpoints but empty nodes remain forever in node knowledge.

We should:
- check that empty node are unloaded when emptied and removed from PRP knownledge
- remove empty nodes during chekpoint from node knowledge too

Remove empty node could have some concurrency problems, it could still be needed from concurrent searches to reach a destination

diegosalvi:
- Improved empty pages checkpoint handling
- removed unused metadata (due to modified empty page handling)
- added mass concurrency delete, checkpoint, read test

[X] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

### DIFF
--- a/herddb-core/src/main/java/herddb/index/blink/BLinkKeyToPageIndex.java
+++ b/herddb-core/src/main/java/herddb/index/blink/BLinkKeyToPageIndex.java
@@ -384,7 +384,14 @@ public class BLinkKeyToPageIndex implements KeyToPageIndex {
          */
         private static final long VERSION_1 = 1L;
 
-        public static final long CURRENT_VERSION = VERSION_1;
+        /**
+         * Removed empty field in BLinkNodeMetadata
+         *
+         * @since 0.9.0
+         */
+        private static final long VERSION_2 = 2L;
+
+        public static final long CURRENT_VERSION = VERSION_2;
 
         private static final long NO_FLAGS = 0L;
 
@@ -419,8 +426,6 @@ public class BLinkKeyToPageIndex implements KeyToPageIndex {
 
                     edos.writeVLong(node.id);
                     edos.writeVLong(node.storeId);
-
-                    edos.writeBoolean(node.empty);
 
                     edos.writeVInt(node.keys);
                     edos.writeVLong(node.bytes);
@@ -458,7 +463,7 @@ public class BLinkKeyToPageIndex implements KeyToPageIndex {
                  */
                 boolean recalculateSize = version == VERSION_0;
 
-                /* flags for future implementations, actually unused (exists from version 1)*/
+                /* flags for future implementations, actually unused (exists from version 1) */
                 @SuppressWarnings("unused")
                 long flags = version > VERSION_0 ? edis.readVLong() : NO_FLAGS;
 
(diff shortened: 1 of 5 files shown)


herddb-core/src/main/java/herddb/index/blink/BLinkKeyToPageIndex.java
```diff
@@ -489,43 +494,109 @@ public Bytes getPosiviveInfinityKey() {
...
+            long outlink = cursor.readZLong();
+            long rightlink = cursor.readZLong();
+
+            boolean hasInf = cursor.readBoolean();
+
+            Bytes rightsep;
+            if (hasInf) {
+                rightsep = Bytes.POSITIVE_INFINITY;
+            } else {
+                rightsep = cursor.readBytesNoCopy();
```
eolivelli: No really sure.
In v1 was already with *readBytesNoCopy*
Should we perform a copy? (In v1 as well)

herddb-utils/src/main/java/herddb/index/blink/BLink.java
```diff
@@ -411,7 +411,22 @@ private boolean attemptUnload(Metadata unload) {
...
-                if (node.empty) {
+                if (node.empty()) {
+
+                    /* If the node existed in policy remove it and unload */
+                    if (policy.remove(node)) {
+                        node.unload(false, false);
+                    }
+
+                    /*
+                     * Remove the node from nodes knowledge: it is a safe operation, if a node is empty it will not used
```
eolivelli: Typo: missing 'be'

herddb-utils/src/main/java/herddb/index/blink/BLink.java
```diff
@@ -411,7 +411,22 @@ private boolean attemptUnload(Metadata unload) {
...
+                     * anymore and it has just an outlink from a possibly real node. If there is a concurrent traversal
+                     * it will continue anyway (nodes reachable though links). Nodes memory is needed only for page
```
eolivelli: Typo 'though'

herddb-utils/src/main/java/herddb/index/blink/BLink.java
```diff
@@ -1771,7 +1780,7 @@ public String toString() {
         /* *** FOR BOTH LEAVES AND NODES *** */
         /* ********************************* */
         boolean empty() {
-            return empty;
+            return outlink != null;
```
eolivelli: I can't understand this.
Is this method used?

diegosalvi: Empty method is used during node search. Empty boolean flag wasn't needed because a node is empty if it was half-merged (and his outlink was set. A node with an outlink is a empty and reclamable node)

herddb-core/src/main/java/herddb/index/blink/BLinkKeyToPageIndex.java
diegosalvi: Interesting, good catch

Issue #347, opened 2019-03-04
Support non null constraints - JDBC support


eolivelli:
With #345 we have introduced the server side handler for NOT-NULL column data types.

Two tasks on the JDBC driver:
1) We have to create end-to-end JDBC tests with NOT-NULL columns, and assert that the client will receive an SQLException.
2) We have to full support the NULLABLE metadata in DatabaseMetaData#getColumns

for 1)
Let's add a simple unit tests which creates a TABLE with all NON NULL datatypes and create tests to trigger the error (one assertionl per column, not only a single INSERT/UPDATE)

for 2)
https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#getColumns(java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String)

see on "herddb-core":
https://github.com/diennea/herddb/blob/master/herddb-core/src/main/java/herddb/core/system/SyscolumnsTableManager.java#L71

here we have to report as NULLABLE = 0:
- all columns in the PK (already implemented)
- all columns with a NON_NULL datatype    (a new ColumnTypes.isNonnullableDataType(int) function is to be created)

see on "herddb-jdbc" module
https://github.com/diennea/herddb/blob/master/herddb-jdbc/src/main/java/herddb/jdbc/HerdDBDatabaseMetadata.java#L826
https://github.com/diennea/herddb/blob/master/herddb-jdbc/src/main/java/herddb/jdbc/HerdDBDatabaseMetadata.java#L869
https://github.com/diennea/herddb/blob/master/herddb-jdbc/src/main/java/herddb/jdbc/HerdDBDatabaseMetadata.java#L950

we have to create tests in JDBC module around getColumns

eolivelli:
cc @amitvc  this is the followup issue of #345

amitvc:
I will start work on this once I close #345

amitvc:
@eolivelli for the UT in jdbc module is it new class or I need to add more tests to existing classes of unit tests?

eolivelli:
@amitvc  this is the task I was referring in the ML

amitvc:
Started work on this.

amitvc:
@eolivelli There are several methods in BasicHerdDBDatasource.java class that are never being used in the code base See example below -
 setWaitForTableSpaceTimeout,getWaitForTableSpace,isDiscoverTableSpaceFromQuery,setDiscoverTableSpaceFromQuery. There are few more.

[...]

PR #365, opened 2019-03-26
Support non null constraints - JDBC support #347

amitvc:
… supported type

### DIFF
--- a/herddb-jdbc/src/test/java/herddb/jdbc/SystemTablesTest.java
+++ b/herddb-jdbc/src/test/java/herddb/jdbc/SystemTablesTest.java
@@ -32,6 +32,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import herddb.model.StatementExecutionException;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
@@ -223,6 +224,23 @@ public class SystemTablesTest {
                         assertFalse(rs.next());
 
                     }
+                }
+            }
+        }
+    }
+
+    @Test(expected = StatementExecutionException.class)
+    public void ensureWeThrowExceptionForNonSupportedNotNullTypes_Double() throws Exception {
+        try (Server server = new Server(new ServerConfiguration(folder.newFolder().toPath()))) {
+            server.start();
+            server.waitForStandaloneBoot();
+            try (HDBClient client = new HDBClient(new ClientConfiguration(folder.newFolder().toPath()));) {
+                client.setClientSideMetadataProvider(new StaticClientSideMetadataProvider(server));
+                try (BasicHerdDBDataSource dataSource = new BasicHerdDBDataSource(client);
+                     Connection con = dataSource.getConnection();
+                     Statement statement = con.createStatement();) {
+                     statement.execute("CREATE TABLE mytable (key string primary key, name string,d1 double not null)");
+
                 }
             }
         }
--- a/herddb-jdbc/src/test/java/herddb/jdbc/SystemTablesTest.java
+++ b/herddb-jdbc/src/test/java/herddb/jdbc/SystemTablesTest.java
@@ -226,145 +226,47 @@ public class SystemTablesTest {
         }
     }
 
-    @Test(expected = SQLException.class)
+    @Test
     public void ensureWeThrowExceptionForNonSupportedNotNullTypes_Double() throws Exception {
         try (Server server = new Server(new ServerConfiguration(folder.newFolder().toPath()))) {
             server.start();
             server.waitForStandaloneBoot();
             try (HDBClient client = new HDBClient(new ClientConfiguration(folder.newFolder().toPath()));) {
                 client.setClientSideMetadataProvider(new StaticClientSideMetadataProvider(server));
-                try (BasicHerdDBDataSource dataSource = new BasicHerdDBDataSource(client);
+                try {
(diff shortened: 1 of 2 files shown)


amitvc:
@eolivelli I added a UT in jdbc/SystemTablesTest.java to demonstrate that we throw the exception when not null constraint is applied for non supported types. Appropriate exception is thrown here - https://github.com/diennea/herddb/blob/master/herddb-core/src/main/java/herddb/server/ServerSideConnectionPeer.java#L705

Currently we dont catch this exception and hence it is not being propagated back to the client over the network. I am guessing we need to handle these things and also interpret them on the client side. Do we have existing protocol or message types defined that we use to communicate these things ?

Exception --
herddb.model.StatementExecutionException: Not null constraints not supported for column type 6
at herddb.model.ColumnTypes.getNonNullTypeForPrimitiveType(ColumnTypes.java:105)
at herddb.sql.DDLSQLPlanner.buildCreateTableStatement(DDLSQLPlanner.java:369)
at herddb.sql.DDLSQLPlanner.plan(DDLSQLPlanner.java:293)
at herddb.sql.DDLSQLPlanner.translate(DDLSQLPlanner.java:270)
at herddb.sql.CalcitePlanner.translate(CalcitePlanner.java:242)
at herddb.server.ServerSideConnectionPeer.handleExecuteStatement(ServerSideConnectionPeer.java:705)

eolivelli:
You are right.
Good catch
It is a regression introduced by the refactoring in which we introduced CompletableFuture in this class.

Please add a try/catch.
In the sql client you should expect a SQLException.

Please check other usages of 'translate' in this class and add tests for them.

amitvc:
@eolivelli - Added more tests for different data types. But these tests are all about table creation. 
Are there already tests for data insertion or updates in the jdbc module where we can add tests for not null constraints on types that are supported and not supported?

herddb-jdbc/src/test/java/herddb/jdbc/SystemTablesTest.java
```diff
@@ -227,4 +225,148 @@ public void test() throws Exception {
...
+    public void ensureWeThrowExceptionForNonSupportedNotNullTypes_BOOLEAN() throws Exception {
+        try (Server server = new Server(new ServerConfiguration(folder.newFolder().toPath()))) {
+            server.start();
+            server.waitForStandaloneBoot();
+            try (HDBClient client = new HDBClient(new ClientConfiguration(folder.newFolder().toPath()));) {
+                client.setClientSideMetadataProvider(new StaticClientSideMetadataProvider(server));
+                try (BasicHerdDBDataSource dataSource = new BasicHerdDBDataSource(client);
+                     Connection con = dataSource.getConnection();
+                     Statement statement = con.createStatement();) {
+                     statement.execute("CREATE TABLE mytable (key string primary key, name string,b1 boolean not null)");
```
eolivelli: please add a test case with PreparedStatement#addBatch and then executeBatch, it will run thru a different code path

herddb-jdbc/src/test/java/herddb/jdbc/SystemTablesTest.java
```diff
@@ -227,4 +225,148 @@ public void test() throws Exception {
...
+                     Connection con = dataSource.getConnection();
+                     Statement statement = con.createStatement();) {
+                    statement.execute("CREATE TABLE mytable (key string primary key, name string,f1 float not null)");
+                }
+            }
+        }
+    }
+
+    @Test(expected = SQLException.class)
+    public void ensureWeThrowExceptionForNonSupportedNotNullTypes_BOOLEAN() throws Exception {
```
eolivelli: we can keep just this one and a test which "works", there is no much value in testing other datatype

eolivelli:
We have to add a test about DatabaseMetadata#getColumns as well, then we are done with this new feature!

thank you very much @amitvc

amitvc:
@eolivelli - Addressed comments and added tests.
Regarding DatabaseMetadata#getColumns I have added some unit test for non null types. I think ColumnTypes.typeToString method for not null types is return text like "{type} not null". In the unit test for the DatabaseMetadata.getColumns i had to assert against that types. I feel that is not correct. If the type is String or NOTNULL_STRING it should still return 'string' as a type. What do you think?

eolivelli:
@amitvc 
these are the specs of getColumns:

https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#getColumns(java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String)

column 5 is
DATA_TYPE  int => SQL type from java.sql.Types 
-> we have to follow the standard java.sql.Types 

column N6  is:
TYPE_NAME String => Data source dependent type name, for a UDT the type name is fully qualified 
-> so we are free to return whatever we want, having 'string not null' would be fine, but I feel it is better to return just "string" 

column N18 is:
IS_NULLABLE String => ISO rules are used to determine the nullability for a column. 
I am expecting here that the result reflect your change

amitvc:
@eolivelli I will address 'string not null' in a separate task since this PR is closed. Please assign me more work or tickets. I am going to play around with herddb with a cluster setup over the weekend.

eolivelli:
@amitvc great.

I am going to cut a release during the weekend or on Monday.
If you perform some smoke tests it will be great.
Ping me if you have problems

amitvc:
@eolivelli I will do some testing locally first thing Monday morning.

Issue #372, opened 2019-04-02
Return appropriate COLUMN_NAME for column types with not null constraints.


amitvc:
Is this a question, a feature request or a bug report?
Its an enhancement.

**QUESTION**

Have you checked our documentation in the GitHub Wiki?
If you could not find an answer there, please consider asking your question on our mailing list at herddb-users@lists.herddb.org.

**FEATURE REQUEST**

1. Please describe the feature you are requesting.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

3. Provide any additional detail on your proposed use case for this feature.

4. If there are some sub-tasks using -[] for each subtask and create a corresponding issue to map to the sub task:
- [ ] [sub-task1-issue-number](example_sub_issue1_link_here): sub-task1 discription here, 
- [ ] [sub-task2-issue-number](example_sub_issue2_link_here): sub-task2 discription here,
- ...

**BUG REPORT**

1. Please describe the issue you observed:

- What did you do?

- What did you expect to see?

- What did you see instead?

aluccaroni:
@amitvc can we close this issue?

amitvc:
Yes

PR #373, opened 2019-04-03
Fix column type message for DatabaseMetaData#getColumns method

amitvc:

### DIFF
--- a/herddb-core/src/main/java/herddb/core/system/SyscolumnsTableManager.java
+++ b/herddb-core/src/main/java/herddb/core/system/SyscolumnsTableManager.java
@@ -45,6 +45,7 @@ public class SyscolumnsTableManager extends AbstractSystemTableManager {
             .column("ordinal_position", ColumnTypes.INTEGER)
             .column("is_nullable", ColumnTypes.INTEGER)
             .column("data_type", ColumnTypes.STRING)
+            .column("type_name", ColumnTypes.STRING)
             .column("auto_increment", ColumnTypes.INTEGER)
             .primaryKey("column_name", false)
             .primaryKey("table_name", false)
@@ -71,7 +72,8 @@ public class SyscolumnsTableManager extends AbstractSystemTableManager {
                         "column_name", c.name,
                         "ordinal_position", pos++,
                         "is_nullable", nonNullCType ? DatabaseMetaData.columnNoNulls : DatabaseMetaData.columnNullable,
-                        "data_type", data_type,
+                        "data_type",  ColumnTypes.sqlDataType(c.type),
+                        "type_name", data_type,
                         "auto_increment", (pk && t.auto_increment)?1:0
                 ));
             }
--- a/herddb-core/src/main/java/herddb/model/ColumnTypes.java
+++ b/herddb-core/src/main/java/herddb/model/ColumnTypes.java
@@ -60,11 +60,11 @@ public class ColumnTypes {
             case BOOLEAN:
                 return "boolean";
             case NOTNULL_STRING:
-                return "string not null";
+                return "string";
             case NOTNULL_INTEGER:
-                return "integer not null";
+                return "integer";
             case NOTNULL_LONG:
-                return "long not null";
+                return "long";
             default:
                 return "type?" + type;
         }
--- a/herddb-core/src/main/java/herddb/model/ColumnTypes.java
+++ b/herddb-core/src/main/java/herddb/model/ColumnTypes.java
@@ -60,11 +60,11 @@ public class ColumnTypes {
             case BOOLEAN:
                 return "boolean";
             case NOTNULL_STRING:
-                return "string";
+                return "string not null";
             case NOTNULL_INTEGER:
-                return "integer";
+                return "integer not null";
             case NOTNULL_LONG:
-                return "long";
+                return "long not null";
             default:
                 return "type?" + type;
         }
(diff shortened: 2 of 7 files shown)


amitvc:
@eolivelli PR for https://github.com/diennea/herddb/issues/372

eolivelli:
@diegosalvi what do you think about this ?
It is better to merge this one before cutting the release.

@amitvc  can you explain better why are you doing this change in ColumnTypes instead of doing in at JDBC Driver level ?

in HerdDB "types" model we have "string not null" and "string"

I can accept this patch but we should take into account that it won't be easy to change this behaviour in the future without breaking compatibility

amitvc:
I see your point @eolivelli. Maybe for now we don't use this patch. I think we don't need to do the change in ColumnTypes.java. I will see where in the jdbc driver I can make this change. Honestly this more of cosmetic change than anything else. I am ok to not use this.

eolivelli:
Ok let's do it only in jdbc driver

amitvc:
@eolivelli I think the DatabaseMetaData resultset field should **TYPE_NAME String => Data source dependent type name, for a UDT the type name is fully qualified**. See doc - https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#getColumns(java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String)

herddb-core/src/main/java/herddb/model/ColumnTypes.java
```diff
@@ -60,11 +60,11 @@ public static String typeToString(int type) {
             case BOOLEAN:
                 return "boolean";
             case NOTNULL_STRING:
-                return "string not null";
+                return "string";
```
eolivelli: So here and for the other not null rypes we have to return 'string not null'

Cc @diegosalvi

herddb-jdbc/src/test/java/herddb/jdbc/SystemTablesTest.java
```diff
@@ -266,7 +266,7 @@ public void ensureWeThrowAppropriateExceptionWhenNotNullConstraintViolated_AddBa
                     insertPs.executeBatch();
                 } catch(SQLException ex) {
                     assertTrue(ex.getMessage().contains("StatementExecutionException"));
-                    assertTrue(ex.getMessage().contains("Cannot have null value in non null type string not null"));
```
eolivelli: we should reword the exception message: `cannot have a null for columns of type 'string not null'`

amitvc:
@eolivelli I have made changes in the jdbc layer for the not null type.

herddb-jdbc/src/main/java/herddb/jdbc/HerdDBDatabaseMetadata.java
```diff
@@ -947,7 +947,7 @@ public ResultSet getTableTypes() throws SQLException {
     @SuppressFBWarnings("SQL_NONCONSTANT_STRING_PASSED_TO_EXECUTE")
     public ResultSet getColumns(String catalog, String schemaPattern, String tableNamePattern, String columnNamePattern) throws SQLException {
 
-        String query = "SELECT table_name,column_name,data_type,auto_increment,is_nullable,ordinal_position FROM SYSCOLUMNS WHERE 1=1 ";
+        String query = "SELECT table_name,column_name,data_type,type_name,auto_increment,is_nullable,ordinal_position FROM SYSCOLUMNS WHERE 1=1 ";
```
eolivelli: are we missing an UT for this change ?
I am expecting some assertion about DatabaseMetadata#getColumns

amitvc:
@eolivelli addressed comments and added UT. The new column definition type 'TYPE_NAME' was missing in the result set. I added that.

herddb-core/src/main/java/herddb/core/system/SyscolumnsTableManager.java
```diff
@@ -71,7 +72,8 @@ public SyscolumnsTableManager(TableSpaceManager parent) {
                         "column_name", c.name,
                         "ordinal_position", pos++,
                         "is_nullable", nonNullCType ? DatabaseMetaData.columnNoNulls : DatabaseMetaData.columnNullable,
-                        "data_type", data_type,
+                        "data_type",  ColumnTypes.sqlDataType(c.type),
+                        "type_name", data_type,
```
diegosalvi: Can you change data_type var to another name? Having a data_type record field that not report data_type var and a type_name that report data_type value is a little misleading

eolivelli: @diegosalvi  "data_type" is already here, we can't change it without breaking compatibility with old clients

eolivelli: @amitvc in this file we already had 'data_type' 
>  String data_type = ColumnTypes.typeToString(c.type);
Is that variable no more used ?

As said, we shouldn't change the behaviour for "data_type", can you explain better your change here ?

diegosalvi: I didn't meant to change "data_type" string but just rename data_type variable to type_name variable (such variable has been created 5-6 lines before and a rename won't break anything)... is just strange assign data_type value to "type_name" string and ColumnTypes.sqlDataType(c.type) to "data_type" string. [Otherwise the next time I'll stumble there there I'll need time to check if is an error or is the expected behaviour :(]

amitvc: Good catch. It's my bad. I should have renamed the variable.

amitvc:
@eolivelli @diegosalvi done.

eolivelli:
Awesome.
Will merge as soon as Travis finished

amitvc:
Thank you.

eolivelli:
Coverall is not very happy that we are not covering one of the new methods
Can you please create an issue or add an UT around ColumnTypes?

https://coveralls.io/builds/22824354/source?filename=herddb-core%2Fsrc%2Fmain%2Fjava%2Fherddb%2Fmodel%2FColumnTypes.java#L126

Btw This pr looks good. Thanks

amitvc:
@eolivelli I will add UT around this tonight.

amitvc:
@eolivelli Added more column types to the UT. This should increase the coverage.

PR #376, opened 2019-04-09
Fix FileCommitLog max sync time configuration from 1 microsecond to 1 millisecond

diegosalvi:
txlog.synctimeout defaulted to 1 and in FileCommitLog was used as microseconds.

Modified to default to 1 millisecond. (Internally FileCommitLog will use nanoseconds to avoid conversions every time)

 - [ X ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)


eolivelli:
good catch !
We should cherry pick to release/0.9 branch

eolivelli:
cherry picked to release/0.9

PR #379, opened 2019-04-19
Make BookKeeper ledgers path configurable and let HerdDBEmbeddedDataSource to start in cluster mode

eolivelli:
- Make BookKeeper ledgers path configurable
- Separate BK zookeeper namespace from HerdDB default one
- Simplify the boot of in clustermode using HerdDBEmbeddedDataSource

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ cache:
 jobs:
   include:
   - stage: test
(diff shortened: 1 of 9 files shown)


eolivelli:
I have to fix a jdbc test

PR #387, opened 2019-05-11
Use LongPoll for tailing read on BookKeeper

eolivelli:
Since BookKeeper 4.5 there is the long-poll read feature, see https://issues.apache.org/jira/browse/BOOKKEEPER-670.

This change introduces long-pool reads and refactors the Follower.

Summary of changes:
- change CommitLog interface and introduce FollowerContext (introduce default methods for implementations that do not support 'tailing')
- move the Follower code to new BookKeeper API (LedgerHandle > ReadHandle)
- keep the current ReadHandle open, this save lots of resources on ZK (each ReadHandle is a watch,
and opening a ledger means queries on ZK and on Bookies)
- do not read continuously the "actual ledgers list" from ZK
- use Long-Poll reads of BookKeeper
- clean up logger usage
- move tests related to replication to a better package name
- upgrade to Netty latest and greatest
- upgrade BookKeeper to 4.9.1

This is only the first patch in the set.
The next patch will focus on using less threads on the Follower in order to be able to "follow" more and more tablespaces with the same resources.

After this patch I can see real improvements on the memory usage of both the bookies and the followers. There is less garbage in general and fewer ZK calls.

### DIFF
--- a/herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
+++ b/herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
@@ -63,6 +63,8 @@ public class BookkeeperCommitLog extends CommitLog {
     private static final Logger LOGGER = Logger.getLogger(BookkeeperCommitLog.class.getName());
     // Max number for entry to read while tailing
     private static final int MAX_ENTRY_TO_TAIL = 5000;
+    // Max time to wait for an entry to arrive
+    private static final int LONG_POLL_TIMEOUT = 1000;
     private final String sharedSecret = "herddb";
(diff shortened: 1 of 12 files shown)


eolivelli:
this is still a WIP, I left a lot of logs in order to debug test cases failures

eolivelli:
@amitvc this is a core change.
You  can take a look and see how we perform replication using BookKeeper.

eolivelli:
tagging @Pols95 as this fix reduces a lot the garbage you reported on https://github.com/diennea/carapaceproxy

eolivelli:
thanks @diegosalvi , merging now

PR #390, opened 2019-05-22
Adding show create table feature

amitvc:
This is the initial commit for the work need for 'Show Create table' command from herddb cli.


amitvc:
@eolivelli Initial commit for this work. I was overthinking this feature and spent way too much time looking at syscolumnsTable etc. Everything was available in the table object itself. The output on the herddb cli side looks like this --
 tabledef                                                                 | 
| --------                                                                 | 
| CREATE TABLE herd.test3(id integer,s1 string,n1 integer,Primary key(id)) | 

| tabledef                                                                                   | 
| --------                                                                                   | 
| CREATE TABLE herd.test4(id integer not null,s1 string not null,n1 integer,Primary key(id)) | 

| tabledef                                                                   | 
| --------                                                                   | 
| CREATE TABLE herd.test2(k1 string,s1 string,n1 integer,Primary key(k1,s1)) |

amitvc:
Couple of observations - 

When I create a table with auto_increment the column and not null the column type in Table object is incorrect. It still shows the base type but it does not recognize it as not null. 
Example 
Create table test (id int not null auto_increment, s1 string, primary key(id))
The column type for id should be 12 but it is 2. 
If I remove auto_increment the column type will be 12.

amitvc:
And still adding unit tests for CalcitePlanner

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
```diff
@@ -341,6 +341,84 @@ public TranslatedQuery translate(String defaultTableSpace, String query, List<Ob
...
+                        )
+                );
+                ExecutionPlan executionPlan = ExecutionPlan.simple(
+                        new SQLPlannedOperationStatement(values),
+                        values
+                );
+                return new TranslatedQuery(executionPlan, new SQLStatementEvaluationContext(query, parameters));
+            }
+        }
+        ValuesOp values = new ValuesOp(manager.getNodeId(),
```
eolivelli: Please throw an error, the client must see an hard error

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
```diff
@@ -341,6 +341,84 @@ public TranslatedQuery translate(String defaultTableSpace, String query, List<Ob
         }
     }
 
+    private TranslatedQuery calculateShowCreateTable(String query, List<Object> parameters) {
+        String items[] = { "SHOW","CREATE", "TABLE"};
+        if(Arrays.stream(items).allMatch(query::contains)) {
+            query = query.substring(Arrays.stream(items).collect(Collectors.joining(" ")).length()).trim();
+            String tableSpace = "herd";
```
eolivelli: The implicit tablespace is the parameter defaultTableSpace above in *translate* method

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
```diff
@@ -341,6 +341,84 @@ public TranslatedQuery translate(String defaultTableSpace, String query, List<Ob
...
+            if (query.contains(".")) {
+                String tokens[] = query.split("\\.");
+                tableSpace = tokens[0].trim();
+                tableName = tokens[1].trim();
+            } else {
+                tableName = query.trim();
+            }
+            try {
+                Table t = manager.getTableSpaceManager(tableSpace).getTableManager(tableName).getTable();
+                if (t != null) {
```
eolivelli: You should also check if the table is bound to an uncommitted transaction. On this case you have to handle it has not existing if we are not inside the scope of the transaction that created the table

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
```diff
@@ -341,6 +341,84 @@ public TranslatedQuery translate(String defaultTableSpace, String query, List<Ob
...
+                    StringBuilder sb = new StringBuilder("CREATE TABLE "+tableSpace+"."+tableName);
+                    StringJoiner joiner = new StringJoiner(",","(", ")");
+                    for(Column c : t.getColumns()) {
+                        joiner.add(c.name + " "+ColumnTypes.typeToString(c.type));
+                    }
+                    if(t.getPrimaryKey().length > 0) {
+                        joiner.add("Primary key("+ Arrays.stream(t.getPrimaryKey()).collect(Collectors.joining(",")) + ")");
+                    }
+                    sb.append(joiner.toString());
```
eolivelli: You are missing 'create index....' in case of indexes on the table

herddb-core/src/test/java/herddb/sql/CalcitePlannerTest.java
```diff
@@ -463,6 +460,76 @@ public void explainPlanTest() throws Exception {
         }
     }
 
+    @Test
+    public void showCreateTableTest() throws Exception {
```
eolivelli: Please also create a test for a table created in a transaction, without committing the transaction

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
```diff
@@ -341,6 +341,84 @@ public TranslatedQuery translate(String defaultTableSpace, String query, List<Ob
...
+                            )
+                    );
+                    ExecutionPlan executionPlan = ExecutionPlan.simple(
+                            new SQLPlannedOperationStatement(values),
+                            values
+                    );
+                    return new TranslatedQuery(executionPlan, new SQLStatementEvaluationContext(query, parameters));
+                }
+
+            }catch (NullPointerException ex) {
```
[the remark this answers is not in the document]
amitvc: It happens when someone queries for a non-existent table. Maybe I should not catch the exception and send the actual error back to the user. But the error in this case would be null pointer exception that occurs Line 358 : CalcitePlanner.java - manager.getTableSpaceManager(tableSpace).getTableManager(tableName)

Could also happen if TableSpace is non-existent. What should be the strategy here? Should we let the user know the table space or table does not exist?

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
amitvc: Makes sense. Will do.

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
amitvc: Good catch. I will pass that to the method so can be used here.

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
amitvc: Good catch. I will add those.

herddb-core/src/test/java/herddb/sql/CalcitePlannerTest.java
amitvc: Yes will do.

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
amitvc: Can you please elaborate on this?

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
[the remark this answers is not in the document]
eolivelli: you can prevent the NullPointerException with a null check and then throw a specific error, in this case TableDoesNotExistException is the best one

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
eolivelli: (sorry yesterday when I sent the first feedback I did not have my laptop but I was using my smartphone).


see TableManager#createdInTransaction field.

If the client is issuing this query:
`SHOW CREATE TABLE tspace1.foo`
and the TableManager for _tspace1.foo_ has createdInTransaction = 1234 this means that that table has been created during transaction 1234 but the transaction still has not been committed yet, so the table should not be "visible" to other transactions.

I am seeing now that CalcitePlanner#translate does not known anything about transactions.

So in this case I prefer to throw a TableNotExistsException in case of createdInTransaction > 0.
This way we won't let other transactions see the table if not committed.
The tradeoff here is that the transaction that is creating the table won't be able to use "SHOW CREATE TABLE", but this is not a big deal, because usually SHOW CREATE TABLE is used by operators in order to inspect the contents of a database.

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
```diff
@@ -280,6 +275,11 @@ public TranslatedQuery translate(String defaultTableSpace, String query, List<Ob
                 );
                 return new TranslatedQuery(executionPlan, new SQLStatementEvaluationContext(query, parameters));
 
+            }
+
+            if(query.startsWith("SHOW")) {
+                return calculateShowCreateTable(query, parameters);
+
```
eolivelli: nit: remove empty space

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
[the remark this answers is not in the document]
amitvc: Sounds good.

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
```diff
@@ -27,6 +27,9 @@
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+
+import herddb.core.*;
```
eolivelli: Please avoid star imports

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
amitvc: Thanks for the explanation. I will make the change.

amitvc:
Ready for review.

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
amitvc: Done. I am sorry about that. I think my IntelliJ does it without asking me about it sometimes.

aluccaroni:
@eolivelli @amitvc usually (on other DBMS, such as mysql) the SHOW CREATE TABLE command will also output the indexes, so I think it's better to keep that "convenction". If you want to have 2 different commands I think it's better this way:

Show create table without indexes XXX
Show create table XXX

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
```diff
@@ -341,6 +350,80 @@ public TranslatedQuery translate(String defaultTableSpace, String query, List<Ob
...
+                throw new TableDoesNotExistException(String.format("Table %s does not exist.", tableName));
+            }
+
+
+            Table t  = tableManager.getTable();
+            if (t == null) {
+                throw new TableDoesNotExistException(String.format("Table %s does not exist.", tableName));
+            }
+
+            StringBuilder sb = new StringBuilder("CREATE TABLE "+tableSpace+"."+tableName);
```
eolivelli: Please create an utility class for this part, that is the builder of the create table string.
This way you can do unit tests without starting a server.
Please also create unit tests for:
Multi column pk
Auto increment
Indexes, brin and hash
Table without pk (heap table)
All data types, even not null ones

amitvc: Great suggestion. I will refactor this.

amitvc:
> @eolivelli @amitvc usually (on other DBMS, such as mysql) the SHOW CREATE TABLE command will also output the indexes, so I think it's better to keep that "convenction". If you want to have 2 different commands I think it's better this way:
> 
> Show create table without indexes XXX
> Show create table XXX

eolivelli:
@amitvc how are you doing with this PR ?
We are going to cut 0.10.0 release next week.
It would be great to include this feature !

amitvc:
@eolivelli I will have this out today.

eolivelli:
Great work!

amitvc:
@eolivelli I have added more Unit tests based on your feedback.

aluccaroni:
@amitvc @eolivelli 
I see that both ShowCreateTableCalculator.java and ServerSideConnectionPeer.java miss the License Header (see licenseheader.txt in the root of the master for the template).
Also, in ShowCreateTableCalculatorTest.java, there is a star import...do we allow that?

eolivelli:
@aluccaroni
Good catch.
I am sorry I had already merged this patch.


Can you please @amitvc follow up with the fix for @aluccaroni comments?

PR #393, opened 2019-06-04
Adding license headers and more Unit test for show create table functionality

amitvc:
- Added support for auto_increment.

- End to end testing with using output of show create table to be used in creation of table.

-  Fix license headers.


amitvc:
@eolivelli  I have merged more UT and fixed the license headers.

PR #398, opened 2019-06-17
Read/Write to ZooKeeper in CLI cluster management commands

eolivelli:
With this change the CLI will read and write direclty to/from ZooKeeper, this way the operator can manage the cluster without having the need of an up and running server in the cluster.


eolivelli:
This is the 
https://github.com/diennea/herddb/wiki/Clustermanagement-cookbook

eolivelli:
This is the guide for the tool (I have only added the '--set-leader' command, any of the other commands were already implemented (by @caiok )

https://github.com/diennea/herddb/wiki/Cluster-Management-Cookbook

cc @diegosalvi @amitvc @aluccaroni

PR #402, opened 2019-06-21
HerdDB Collections Framework

eolivelli:
This is a proof of concept for a  "Memory Bounded Collections"  Framework based on HerdDB.
We are going to exploit the instrinsic capacity of an HerdDB server to put a bound on the total amount of memory retained in the heap.

This patch introduces a new module herddb-collections with the first implementation of a "TmpMap".
The patch also cleans up a a bunch of usages of loggers at level "SEVERE", downgrading to "INFO".
The patch introduces on DBManager the ability to not use an additional ThreadPool for callbacks, as we want HerdDB Collections to not need additional "threads" (apart from the db-activator one) in order to work.
The patch also fixes an  unused constructor of SimpleByteArrayInputStream that we are not starting to use.

As this framework is intended to be a replacement of powerfull www.mapdb.org project inside MagNews code base for a _specific workload_ this patch includes a minimal demonstrative test case that asserts that  **HerdDB Collections outperforms MapDB** in the relevant workload.

An example usage is:
```

    @Test
    public void testIntMap() throws Exception {
        try (CollectionsManager manager = CollectionsManager
                .builder()
                .maxMemory(10 * 1024 * 1024)
                .tmpDirectory(tmpDir.newFolder().toPath())
                .build()) {
            manager.start();
            try (TmpMap<Integer, String> tmpMap = manager.createIntKeyedMap()) {
                for (int i = 0; i < 1000; i++) {
                    tmpMap.put(i, "foo" + i);
                }
                for (int i = 0; i < 1000; i++) {
                    assertTrue(tmpMap.containsKey(i));
                }
                for (int i = 0; i < 1000; i++) {
                    assertEquals("foo" + i, tmpMap.get(i));
                }
            }
        }
    }


```


herddb-collections/pom.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>org.herddb</groupId>
+        <artifactId>herddb-parent</artifactId>
+        <version>0.10.0-SNAPSHOT</version>
+        <relativePath>..</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <name>HerdDB JDBC Collections</name>
```
eolivelli: Typo, this project has nothing to do with jdbc

herddb-collections/pom.xml
```diff
@@ -0,0 +1,23 @@
...
+    <artifactId>herddb-collections</artifactId>
+    <packaging>jar</packaging>
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>herddb-core</artifactId>
+            <version>${project.version}</version>
```
eolivelli: Todo: exclude any dependency not useful for the clients, like bookkeeper or zookeeper or netty...

herddb-collections/src/main/java/herddb/collections/CollectionsManager.java
```diff
@@ -0,0 +1,331 @@
...
+        server = new DBManager("localhost",
+                new MemoryMetadataStorageManager(),
+                new FileDataStorageManager(tmpDirectory,
+                        tmpDirectory, 0, false /* fsync */,
+                        false /* o_direct */, false /* o_direct */, NullStatsLogger.INSTANCE),
+                new MemoryCommitLogManager(), tmpDirectory,
+                new ServerHostData("localhost", 0, "", false, Collections.emptyMap()),
+                configuration, NullStatsLogger.INSTANCE);
+        server.setMaxDataUsedMemory(maxMemory);
+        valuesSerializer = new Serializer() {
```
eolivelli: This is only a POC it should be configurable

herddb-collections/src/main/java/herddb/collections/CollectionsManager.java
```diff
@@ -0,0 +1,331 @@
...
+            return this;
+        }
+
+        public CollectionsManager build() {
+            return new CollectionsManager(maxMemory, tmpDirectory);
+        }
+
+    }
+
+    public <V> TmpMap<Integer, V> createIntKeyedMap() {
```
eolivelli: This is only a POC, we should have specific classes with 'int' instead of 'Integer' in order to save allocations

herddb-collections/src/main/java/herddb/collections/CollectionsManager.java
```diff
@@ -0,0 +1,331 @@
...
+        final TableSpaceManager tableSpaceManager = server.getTableSpaceManager(TableSpace.DEFAULT);
+        final String tmpTableName = "tmp" + tableNameGenerator.incrementAndGet();
+        Table table = Table
+                .builder()
+                .name(tmpTableName)
+                .column("pk", ColumnTypes.INTEGER)
+                .column("v", ColumnTypes.BYTEARRAY)
```
eolivelli: Todo: this table schema is not really needed
As we are using the low level APIs which deal only with pairs of byte[]

eolivelli:
@aluccaroni  PTAL

herddb-collections/src/main/java/herddb/collections/CollectionsManager.java
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+    }
+
+    public class TmpMapBuilder<V> {
+
+        private ValueSerializer<V> valueSerializer = DEFAULT_VALUE_SERIALIZER;
+        private int expectedValueSize = 64;
+
+        public class IntTmpMapBuilder<VI extends V> {
```
amitvc: Do we need to add LongTmpMapBuilder? We support NOTNULL_LONG key type.

herddb-collections/src/main/java/herddb/collections/TmpMap.java
```diff
@@ -0,0 +1,104 @@
...
+ * <li>Perform lookups
+ * <li>Close the map
+ * </ul>.
+ * <p>
+ * <b>The map is not expected to be thread safe. Do not wrap it with an external synchronization mechanism as it is
+ * anyway sharing internal data structures with every other collection allocated by the same CollectionsManager</b>
+ *
+ * @author enrico.olivelli
+ */
+public interface TmpMap<K, V> extends AutoCloseable {
```
amitvc: Should there be support for things like computeIfAbsent, putifAbsent

herddb-collections/src/main/java/herddb/collections/CollectionsManager.java
eolivelli: This is the first stub for this new sub project.
I guess we will add support for 'long' soon.
I can do it now if you think it is useful.
In the cases we will cover in MagNews we only have ints or Pojos.
Supporting longs will be very easy.
I will do.
I feel we won't ever support timestamps, we are not going to create a time series DB

herddb-collections/src/main/java/herddb/collections/TmpMap.java
eolivelli: Good idea
We can do it in the future.
If you think about a non thread safe implementation you can implement it very easily.
But if we go thru a thread safe implementation we need to be careful for computeXxxx.

putIfAbsent is easy, it is like current put() but without the UPDATE.

Actually I don't need such features.
I prefer to add features only if needed: once you release a public API you must maintain it for long term

eolivelli:
I also have to fix a test case
CreateTableTest.weThrowExceptionWhenNullableDataTypeByteArrayUsedasPk Expected exception: java.lang.IllegalArgumentException

eolivelli:
@amitvc I have addressed all of your comments.

I have also dropped MapDb.org example as in order to have it we had to keep the dependency in the pom.xml file and it still imports a lot of dependencies

herddb-collections/src/main/java/herddb/collections/TmpMapImpl.java
```diff
@@ -0,0 +1,225 @@
...
+    @Override
+    public void remove(K key) throws Exception {
+        StatementEvaluationContext context = new PutStatementEvaluationContext(key, null);
+        tableSpaceManager.executeStatement(delete, context, TransactionContext.NO_TRANSACTION);
+
+    }
+
+    @Override
+    public boolean isSwapped() {
+        return stats.getLoadedPagesCount() > 0
```
diegosalvi: Why swapped == loaded > 0 || unloaded > 0? Shouldn't just be  unloaded > 0?

diegosalvi:
Branch merged, remaining work on putIfAbsent and similars will be added as soon as needed.

PR #403, opened 2019-06-27
Fix a problem with the local client side cache in case of server restart

eolivelli:


eolivelli:
This is to be cherry picked to release/0.10 and we should release an hotfix version as soon as possible

diegosalvi:
Nice and simple solution

PR #410, opened 2019-07-15
HerdDB Collections Framework enhancements:

eolivelli:
- clean up dependencies
- disable the SQL Parser
- disable JMX
- add the ability to override HerdDB Configuration


herddb-core/src/main/java/herddb/server/ServerConfiguration.java
```diff
@@ -245,11 +245,10 @@
     public static final String PROPERTY_JMX_ENABLE = "server.jmx.enable";
     public static final boolean PROPERTY_JMX_ENABLE_DEFAULT = true;
 
-    public static final String PROPERTY_PLANNER_TYPE = "server.planner.type";
-    public static final String PLANNER_TYPE_JSQLPARSER = "jsqlparser";
```
eolivelli: This option is not working anymore since many major versions, it worked only during the switch to Calcite

PR #411, opened 2019-07-18
Use Context Class Loader for default serializer

eolivelli:
By default ObjectInputStream does not take into consideration the Thread Context Class Loader and this may lead to unespected beheaviour in clients of the Collections Framework.
With this change the default value deserializer will use the  Thread Context Class Loader


eolivelli:
The problem has been reported by @carlam16

PR #413, opened 2019-07-18
Create a Script to start a 3-node clusers (without docker)

eolivelli:
Create a Script to start a 3-node clusers (without docker)
Fix a typo in the default server.properties file


eolivelli:
please consider only the latest commit while reviewing this patch, it is based on #412

PR #416, opened 2019-07-25
Fix flush new pages during checkpoint with a concurrent flush due to unload

diegosalvi:
If a new page flushed during checkpoint is concurrently flushed by page replacement policy for unload a memory leak could happen with a page not known anymore to PRP but still loaded and known to TableManager

### DIFF
--- a/herddb-core/src/main/java/herddb/core/TableManager.java
+++ b/herddb-core/src/main/java/herddb/core/TableManager.java
@@ -743,9 +743,8 @@ public final class TableManager implements AbstractTableManager, Page.Owner {
         if (flushed) {
             /*
              * Replace the page in memory with his immutable version (faster modification checks). We can
-             * replace the page even if we didn't flush it (i.e.:flushed by some other thread) because we are
-             * running during a checkpoint and no other page write could happen (thus our page copy is fully
-             * equivalent to really flushed one).
+             * replace the page with the immutable one from memory running during a checkpoint and no other
+             * page write could happen (thus our page copy is fully equivalent to really flushed one).
              *
              * We replace the page only if we have actually flushed it.. if it was flushed by another thread
              * it was flushed due to unload request (no concurrent checkpoints) and then removed from pages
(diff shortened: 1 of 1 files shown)


herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -740,17 +740,26 @@ public void unload(long pageId) {
...
+             *
+             * We replace the page only if we have actually flushed it.. if it was flushed by another thread
+             * it was flushed due to unload request (no concurrent checkpoints) and then removed from pages
+             * (or the thread is going to remove it). We don't want to keep knowledge of a page not know
+             * anymore to page replacement policy (It can't be remove again)
+             *
+             * For similar reason we replace the page only if there actually is a page in the first place. If
+             * a concurrent thread flushed and removed the page we don't want to add it again.
+             */
+            pages.computeIfPresent(page.pageId, (i,p) -> page.toImmutable());
```
eolivelli: ` pages.computeIfPresent(page.pageId, (i,p) -> page.toImmutable());`

better form:
` pages.computeIfPresent(page.pageId, (i,p) -> p.toImmutable());`

in fact your first form creates an inner class (or something like that) that needs to "capture" the context,
the second form is a simple function with two arguments

diegosalvi:
It is a fix because adding the page without check can add a page back again in pages collection after another thread already removed it from page replacement policy and pages collections.
It remain a "phantom" page not memory traked. It won't ever be unloaded again if not cleaned from some checkpoint.

herddb-core/src/main/java/herddb/core/TableManager.java
diegosalvi: Practically speaking we could switch between the 2 implementations but can be p != page in some unfortunate cases. (But the should be fully equivalent).

I'll chage the  implementation and add an assertion to enforce it

diegosalvi: On second thought... adding an assertion require to have both p and page thus creating a "capturing" inner class

herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -740,17 +740,26 @@ public void unload(long pageId) {
...
             LOGGER.log(Level.INFO, "New page {0} already flushed in a concurrent thread", page.pageId);
         }
 
-        /*
-         * Replace the page in memory with his immutable version (faster modification checks). We can
-         * replace the page even if we didn't flush it (i.e.:flushed by some other thread) because we are
-         * running during a checkpoint and no other page write could happen (thus our page copy is fully
-         * equivalent to really flushed one)
-         */
-        pages.put(page.pageId, page.toImmutable());
```
eolivelli: Sorry, I missed this part 
put -> computeIfPresent

Issue #426, opened 2019-07-30
NPE on setPageDirty


diegosalvi:
**BUG REPORT**

Got a NPE on a clean installed standalone HerdDB

```
WARNING: Exception while updating message status: java.util.concurrent.CompletionException: java.lang.NullPointerException
        at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:314)
        at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:319)
        at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:645)
        at java.base/java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:478)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.lang.NullPointerException
        at herddb.core.PageSet.setPageDirty(PageSet.java:109)
        at herddb.core.PageSet.setPageDirty(PageSet.java:119)
        at herddb.core.TableManager.applyUpdate(TableManager.java:1675)
        at herddb.core.TableManager.onTransactionCommit(TableManager.java:1413)
        at herddb.core.TableSpaceManager.apply(TableSpaceManager.java:361)
        at herddb.core.TableSpaceManager.lambda$commitTransaction$16(TableSpaceManager.java:1692)
        at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:642)
        ... 5 more
```

diegosalvi:
It appear that selected page isn't listed in PageSet.activePages:
* or PK returned a wrong unexisting page
* or the page wasn't added to active pages but it should have been added
* or the page was removed from active pages but it souldn't

PR #428, opened 2019-07-30
Add more logs for possible NPE on unknown page set dirty (#426)

diegosalvi:
Add more logs for possible NPE on unknown page set dirty (#426)
Fixed old record retrieval on applyDelete to match retrieval of applyUpdate (and add the same logs too)

### DIFF
--- a/herddb-core/src/main/java/herddb/core/TableManager.java
+++ b/herddb-core/src/main/java/herddb/core/TableManager.java
@@ -1561,7 +1561,8 @@ public final class TableManager implements AbstractTableManager, Page.Owner {
         /* This could be a normal or a temporary modifiable page */
         final Long pageId = keyToPage.remove(key);
         if (pageId == null) {
-            throw new IllegalStateException("corrupted transaction log: key " + key + " is not present in table " + table.name);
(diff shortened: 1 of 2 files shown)


herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -1584,18 +1585,26 @@ private void applyDelete(Bytes key) throws DataStorageManagerException {
         if (indexes == null) {
             /* We don't need the page if isn't loaded or isn't a mutable new page */
             page = newPages.get(pageId);
-            previous = null;
 
             if (page != null) {
                 pageReplacementPolicy.pageHit(page);
+                previous = page.get(key);
```
eolivelli: this is a behavior change. this is not only logs

diegosalvi: Yes I know. I cant split the commit if needed. This code part should be equal to update one. It's about old record retrieval to update dirty status as precise as possible without load data if there isn't

eolivelli: ok

PR #430, opened 2019-07-31
Primary Key index files are not deleted after a DROP TABLE

eolivelli:


herddb-collections/src/main/java/herddb/collections/CollectionsManager.java
```diff
@@ -145,7 +145,6 @@ private CollectionsManager(long maxMemory, Path tmpDirectory, Properties additio
                 new MemoryCommitLogManager(false /*serialize*/), tmpDirectory,
                 new ServerHostData("localhost", 0, "", false, Collections.emptyMap()),
                 configuration, NullStatsLogger.INSTANCE);
-        server.setMaxDataUsedMemory(maxMemory);
```
eolivelli: this was a bug, it made difficult to override memory settings

PR #436, opened 2019-08-06
Do not rollback transactions on client socket error

eolivelli:
In 11.0.0 when a client channel is closed every transaction that originated from it is automatically rollback back, this comes with many issues:
- the client opens multiple sockets and the socket is not bound to the transaction
- the rollback will appear in any point in the history and the client actually continues to use the transaction via the other available sockets or opening a new onw: (BEGIN-TX, ROLLBACK-TX, UPDATE-RECORD-WITH-TX, this will break many assumptions
- the rollback may release record level locks but the client is actually still using the connection

the proposal is to have a new simpler "Abandoned transaction" mechanism that rollsback transactions after a very long period of inactivity.
The tradeoff is that with this new way when a client crashes its own transactions are not immediately rollback back.


herddb-core/src/main/java/herddb/core/TableSpaceManager.java
```diff
@@ -701,15 +704,32 @@ public boolean equals(Object obj) {
...
+        long abandonedTransactionTimeout = now - timeout;
+        for (Transaction t : transactions.values()) {
+            if (t.isAbandoned(abandonedTransactionTimeout)) {
+                LOGGER.log(Level.SEVERE, "forcing rollback of abandoned transaction {0},"
+                        + " created locally at {1},"
+                        + " last activity locally at {2}",
+                        new Object[]{t.transactionId,
+                            new java.sql.Timestamp(t.localCreationTimestamp),
+                            new java.sql.Timestamp(t.lastActivityTs)});
+                long lockStamp = acquireReadLock("forceRollback" + t.transactionId);
```
eolivelli: checkpoint lock

herddb-core/src/main/java/herddb/server/ServerSideConnectionPeer.java
```diff
@@ -103,8 +103,7 @@
     /**
      * Open scanners. The ID is generated by the client
      */
-    private final ConcurrentMap<Long, ServerSideScannerPeer> scanners = new ConcurrentHashMap<>();
-    private final ConcurrentMap<RawString, Set<Long>> openTransactions = new ConcurrentHashMap<>();
```
eolivelli: this was the bad mechanism that was binding a transaction to the originating socket (ServerSideConnectionPeer)

herddb-core/src/test/java/herddb/server/SimpleClientServerTest.java
```diff
@@ -469,22 +489,127 @@ public void testSQLIntegrityViolation() throws Exception {
...
                     }
                 }
 
             }
         }
     }
+    private static final Logger LOG = Logger.getLogger(SimpleClientServerTest.class.getName());
+
+    @Test
+    public void testClientCloseOnConnectionAndResumeTransaction() throws Exception {
```
eolivelli: this is the test about the problems in production

herddb-core/src/test/java/herddb/server/SimpleClientServerTest.java
```diff
@@ -469,22 +489,127 @@ public void testSQLIntegrityViolation() throws Exception {
...
+                        "INSERT INTO mytable (id,s1) values(?,?)", tx,
+                        false, true,
+                        Arrays.asList(2, "test1"));
+
+            }
+        }
+    }
+
+    @Test
+    public void testClientAbandonedTransaction() throws Exception {
```
eolivelli: this is a minimal test for the new feature

diegosalvi:
Really good catch

Il Mar 6 Ago 2019, 12:26 Coveralls <notifications@github.com> ha scritto:

eolivelli:
@diegosalvi we are going to cherry pick and release 0.11.5 as soon as possible

amitvc:
Nice catch on this @eolivelli

PR #438, opened 2019-08-07
Handle write errors on BookKeeper more gracefully

eolivelli:
In case of invalid "Writer" due to a temporary error (not "fencing") we have to create a new ledger in order to recover gracefully from temporary errors on ZooKeeper or on BookKeeper

### DIFF
--- a/herddb-core/src/test/java/herddb/cluster/BookkeeperFailuresTest.java
+++ b/herddb-core/src/test/java/herddb/cluster/BookkeeperFailuresTest.java
@@ -21,6 +21,7 @@ package herddb.cluster;
 
 import static herddb.core.TestUtils.scan;
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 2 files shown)


herddb-core/src/test/java/herddb/cluster/BookkeeperFailuresTest.java
```diff
@@ -469,18 +529,103 @@ public void testBookieNotAvailableDuringTransaction() throws Exception {
...
+                    TransactionContext.NO_TRANSACTION);
 
             try (DataScanner scan = scan(server.getManager(), "select * from t1", Collections.emptyList());) {
                 assertEquals(1, scan.consume().size());
             }
         }
     }
 
+    @Test
+    public void testLedgerClosedError() throws Exception {
```
eolivelli: this is the new test

PR #440, opened 2019-08-08
Count activities on transactions and wait for every activity to complete before starting a rollback or commit

eolivelli:
In case of a timeout any usual JDBC client receives an SQLException and it rollbacks the transaction.
The command (say, an huge DELETE) still continues its execution and so we fall into a case in which a transaction command (ROLLBACK, but the same could happen for a commit) that runs concurrently with an operation pertaining to the same transaction.

We must not fall into this case.

So the fix here is to have some kind of "reference count" of the transaction and wait for any pending activity to complete before proceeding with a rollback or a commit.

Summary of changes:
- add a "refcount" in Transaction class
- bind the DataScanner to the Transaction, this way we can track the end of a SELECT
- wrap the executions of DML in a way such that we track the end of the DML command

### DIFF
--- a/herddb-core/src/main/java/herddb/core/DBManager.java
+++ b/herddb-core/src/main/java/herddb/core/DBManager.java
@@ -695,9 +695,9 @@ public class DBManager implements AutoCloseable, MetadataChangeListener {
             if (plan.mainStatement instanceof ScanStatement) {
                 DataScanner result = scan((ScanStatement) plan.mainStatement, context, transactionContext);
                 // transction can be auto generated during the scan
-                transactionContext = new TransactionContext(result.transactionId);
+                transactionContext = new TransactionContext(result.getTransactionId());
                 return CompletableFuture
-                        .completedFuture(executeDataScannerPlan(plan, result, context, transactionContext));
+                        .completedFuture(new ScanResult(transactionContext.transactionId, result));
 
             } else {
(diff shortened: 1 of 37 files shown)


herddb-core/src/main/java/herddb/core/DBManager.java
```diff
@@ -711,41 +711,6 @@ public StatementExecutionResult executePlan(ExecutionPlan plan, StatementEvaluat
         }
     }
 
-    private StatementExecutionResult executeDataScannerPlan(ExecutionPlan plan, DataScanner result,
```
eolivelli: dead code

eolivelli:
JDBC tests are failing, I will fix them as soon as possible

eolivelli:
This patch is good to go.
But we have to follow up with a fix on the handling of how we close DataScanner instances in ServerSideConnectionPeer.
This is the same problem as for #436 : we are binding the DataScanner to the ServerSideConnectionPeer (socket) that originated it but this was good only in first versions of HerdDB.
Now that the client opens multilple sockets to the same server we should not to such binding.

I guess no one ever noticed that problem because usually users perform look ups by PK and do not request huge result sets

eolivelli:
@diegosalvi  @amitvc 
Travis passed.
I will use this work as base for the next patch.
PTAL soon

PR #443, opened 2019-08-10
Fix BRIN blocks successor of themselves

eolivelli:
Due to a wrong assumption, ordering between blocks tree and blocks next relationship differs when splitting a block and the split key is equals to the next block min key. When traversing blocks tree the old one comes before the new (wrong!) but following nex relationship it follow in the right way.
Due to this misbehaviour can even happens that after a checkpoint or node merges a node next elements is itself (which is deeply wrong)

-----------------------------------------------------------

@diegosalvi  I have created a reproducer for the problem about BRIN.
Now finding a fix should be easier

The test is using a fixed seed for the Random instance so the behaviour is 100% determistic. no need for concurrency

### DIFF
--- a/herddb-core/src/main/java/herddb/index/brin/BRINIndexManager.java
+++ b/herddb-core/src/main/java/herddb/index/brin/BRINIndexManager.java
@@ -109,7 +109,7 @@ public class BRINIndexManager extends AbstractIndexManager {
         }
 
         void serialize(ExtendedDataOutputStream out) throws IOException {
-            out.writeVLong(2); // version
+            out.writeVLong(3); // version
             out.writeVLong(0); // flags for future implementations
 
             out.writeVInt(this.type);
@@ -135,12 +135,12 @@ public class BRINIndexManager extends AbstractIndexManager {
                         if (!head) {
                             out.writeArray(md.firstKey);
                         }
-                        out.writeVLong(md.blockId);
+                        out.writeZLong(md.blockId);
                         out.writeVLong(md.size);
                         out.writeVLong(md.pageId);
 
                         if (!tail) {
-                            out.writeVLong(md.nextBlockId);
+                            out.writeZLong(md.nextBlockId);
                         }
                     }
                     break;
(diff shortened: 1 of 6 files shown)


eolivelli:
@amitvc

diegosalvi:
I've found the problem on BRIN checkpoints. It is caused on the false assumpion that block with BlockStartKey{N,K} will alwayfollow block with BlockStartKey{N,K-1} being K the blockid ever increasing number.

Such assumptions fails in one case and is really wrong:

Head block 100% full with values 1,2,2 (BlockStartKey{HEAD,0}), next block not full with value 2,3 (BlockStartKey{2,1}).
Put 1:  HEAD -> 1,1,2,2 -> split
New chain will be: BlockStartKey{HEAD,0} with 1,1 -> BlockStartKey{2,2} with 2,2 -> BlockStartKey{2,1} with 2,3

In this case ordering from tree and "next chain" differs: tree is ordered by comparator and respect blockid odering on same minKey values but "next chain" doesn't. Right ordering is "next chain" one. Using the previous example and reordering with comparators would generate this wrong ordering:
BlockStartKey{HEAD,0} with 1,1 -> BlockStartKey{2,1} with 2,**3** -> BlockStartKey{2,2} with 2,2

diegosalvi:
Added the fix for brin problem handling BlockStartKey ordering in the right way

diegosalvi:
Yes it will force the update but we cannot trust what is indexed currently.

Unfortunately I got another issue running locally more tests, I'm looking for address it too (always related to a node child of itsef)

diegosalvi:
Remaining problem was related to current block id evaluation after index reload. Fixed computation and added a specific test.

diegosalvi:
CI failed the submission to Coveralls. All test passed

eolivelli:
Travis failed

eolivelli:
Okay.
LGTM

I can't approve thus PR because I have created it.
Please merge and cherry pick

PR #454, opened 2019-09-03
Fix transactions refcounts for aggregating queries

eolivelli:
Fix ref count during aggregations and in general on Scanners.

SimpleDataScanner and StreamingDataScanner are the root holders of references to the Transaction.
Fix refcount and close() methods of every DataScanner implementation.
Prevent "double release" in root DataScanners

### DIFF
--- a/herddb-core/src/test/java/herddb/server/SimpleClientScanTest.java
+++ b/herddb-core/src/test/java/herddb/server/SimpleClientScanTest.java
@@ -1,23 +1,22 @@
 /*
- Licensed to Diennea S.r.l. under one
- or more contributor license agreements. See the NOTICE file
- distributed with this work for additional information
- regarding copyright ownership. Diennea S.r.l. licenses this file
- to you under the Apache License, Version 2.0 (the
- "License"); you may not use this file except in compliance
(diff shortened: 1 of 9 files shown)


herddb-core/src/test/java/herddb/server/SimpleClientScanTest.java
```diff
@@ -247,36 +270,37 @@ private void checkCloseResultSetNotFullyScanned(
         }
     }
 
-    private void checkCloseResultSetNotFullyScannedWithTransactionAndAggregation(final HDBConnection connection, final Server server, boolean withTransaction) throws HDBException, ClientSideMetadataProviderException, InterruptedException {
+    private void checkCloseResultSetNotFullyScannedWithTransactionAndAggregation(final HDBConnection connection,
```
eolivelli: this test case was already there, but it was not called, so this part was not tested.

eolivelli:
@diegosalvi this fix is for master branch only I would prefer not to cherry pick to stable branch 0.11
WDYT ?

eolivelli:
Superseded by #455

PR #455, opened 2019-09-03
Fix transactions refcounts for aggregating queries

eolivelli:
Fix transactions refcounts for aggregating queries
Ensure ResultSets are closed from JDBC Driver
With this patch we are now clearly stating that when you commit, rollback or close a transaction you are implicitly closing open every result set (DataScanner).
The behaviour of previous versions was undefined.
This change in particular affects only scans with more records that the fetch size, this was very unlikely to happen in the original usecase of HerdDB (only PK get/put/delete), in fact if fetchsize if greater than the number of records than all of the records are returned to the client and the DataScanner is automatically closed after sending the results to the client.

### DIFF
--- a/herddb-core/src/main/java/herddb/core/SimpleDataScanner.java
+++ b/herddb-core/src/main/java/herddb/core/SimpleDataScanner.java
@@ -25,6 +25,7 @@ import herddb.model.DataScannerException;
 import herddb.model.Transaction;
 import herddb.utils.DataAccessor;
 import java.util.Iterator;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * Simple data scanner on a in memory MaterializedRecordSet
@@ -37,6 +38,7 @@ public class SimpleDataScanner extends DataScanner {
     private Iterator<DataAccessor> iterator;
     private DataAccessor next;
     private boolean finished;
+    private final AtomicBoolean closed = new AtomicBoolean(false);
 
     public SimpleDataScanner(Transaction transaction, MaterializedRecordSet recordSet) {
         super(transaction, recordSet.fieldNames, recordSet.columns);
@@ -49,6 +51,11 @@ public class SimpleDataScanner extends DataScanner {
 
     @Override
     public void close() throws DataScannerException {
+        if (closed.compareAndSet(false, true)) {
+            if (transaction != null) {
+                transaction.decreaseRefCount();
+            }
+        }
         finished = true;
         try {
             recordSet.close();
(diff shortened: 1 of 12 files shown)


eolivelli:
Supersedes #454

herddb-core/src/main/java/herddb/core/SimpleDataScanner.java
```diff
@@ -49,6 +51,11 @@ public SimpleDataScanner(Transaction transaction, MaterializedRecordSet recordSe
 
     @Override
     public void close() throws DataScannerException {
+        if (closed.compareAndSet(false, true)) {
+            if (transaction != null) {
+                transaction.decreaseRefCount();
+            }
+        }
         finished = true;
```
diegosalvi: Can't be added the whole closing method under close.compareAndSet?
And practically speaking: could we just use an atomiboolean for both "closed" and  "finished" (and "fast closing" the scanner and result set upon recordSet finishing?)

herddb-core/src/main/java/herddb/core/StreamDataScanner.java
```diff
@@ -72,7 +74,11 @@ public DataAccessor next() throws DataScannerException {
 
     @Override
     public void close() throws DataScannerException {
+        if (closed.compareAndSet(false, true)) {
+            if (transaction != null) {
+                transaction.decreaseRefCount();
+            }
+        }
         super.close();
```
diegosalvi: Can we invoke even super.close() under closed.compareAndSet?

herddb-core/src/main/java/herddb/model/DataScanner.java
```diff
@@ -82,9 +82,6 @@ public void forEach(Consumer<DataAccessor> consumer) throws DataScannerException
 
     @Override
     public void close() throws DataScannerException {
-        if (transaction != null) {
```
diegosalvi: I like this: it didn't increment the count it shouldn't decrement either

herddb-core/src/main/java/herddb/core/SimpleDataScanner.java
eolivelli: "fast closing" the scanner and result set upon recordSet finishing? -> this may be lead to unpredictable behaviour.

We are closing a "finished" resultset in ServerSideConnectionPeer and this is part of the "protocol": if there are no more records the ResultSet is closed, and the client won't send the "CLOSE SCANNER" message (this is the most common case for findByPk for instance)

eolivelli: btw I have moved the block inside the critical section

herddb-core/src/main/java/herddb/core/StreamDataScanner.java
eolivelli: done

PR #457, opened 2019-09-05
Better error in case of unreachable server with client in standalone mode

eolivelli:


herddb-core/src/main/java/herddb/server/StaticClientSideMetadataProvider.java
```diff
@@ -55,4 +55,9 @@ public String getTableSpaceLeader(String tableSpace) throws ClientSideMetadataPr
         return nodeId;
     }
 
+    @Override
+    public void requestMetadataRefresh(Exception error) throws ClientSideMetadataProviderException {
+        throw new ClientSideMetadataProviderException("not working in cluser mode. metadata refresh is not possible", error);
```
eolivelli: this is actually the  fix

eolivelli:
@diegosalvi this is good to go now, Travis Passed

eolivelli:
This is the not right approach.

PR #460, opened 2019-09-06
Add extra recovery tests with BookKeeper commit log

eolivelli:


herddb-core/src/main/java/herddb/core/TableSpaceManager.java
```diff
@@ -268,6 +268,7 @@ void recover(TableSpace tableSpaceInfo) throws DataStorageManagerException, LogN
         dataStorageManager.loadTransactions(logSequenceNumber, tableSpaceUUID, t -> {
             transactions.put(t.transactionId, t);
             LOGGER.log(Level.FINER, "{0} {1} tx {2} at boot", new Object[]{nodeId, tableSpaceName, t.transactionId});
+            LOGGER.log(Level.SEVERE, "{0} {1} tx {2} at boot seq {3}", new Object[]{nodeId, tableSpaceName, t.transactionId, t.lastSequenceNumber});
```
diegosalvi: Log level too high. Other line should be removed too

eolivelli: removed

PR #465, opened 2019-09-16
Force full data download on follower recovery with missing ledgers

diegosalvi:
When starting recovery on BookkeeperCommitLog a full recovery is needed if some ledgers from current actualLedgersList doesn't exist anymore

### DIFF
--- a/herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
+++ b/herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
@@ -38,6 +38,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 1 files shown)


herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
```diff
@@ -386,6 +387,21 @@ public void recovery(
                     + ": Local data is absent, and actual ledger list " + this.actualLedgersList.getActiveLedgers()
                     + " does not contain first ledger of ever: " + this.actualLedgersList.getFirstLedger()));
         }
+        for (long ledgerId : actualLedgersList.getActiveLedgers()) {
+            try {
+                bookKeeper.getLedgerManager().readLedgerMetadata(ledgerId).get();
```
eolivelli: is use use org.apache.bookkeeper....FutureUtils.result() you won't have to deal with "ExecutionException"

PR #472, opened 2019-09-24
Right index restore behaviour

diegosalvi:
During tablespace recovery index arent correctly handled and restored again but left unrestored.
Otherwise when using Backup recovery index are restored correctly

### DIFF
--- a/herddb-core/src/main/java/herddb/file/FileDataStorageManager.java
+++ b/herddb-core/src/main/java/herddb/file/FileDataStorageManager.java
@@ -944,6 +944,25 @@ public class FileDataStorageManager extends DataStorageManager {
             }
 
         } catch (IOException err) {
+            LOGGER.log(Level.SEVERE, "Failed to write on path: {0}", pageFile);
+            Path path = pageFile;
+            boolean exists;
+            do {
+                exists = Files.exists(path);
+
+                if (exists) {
+                    LOGGER.log(Level.SEVERE,
+                            "Path {0}: directory {1}, file {2}, link {3}, writable {4}, readable {5}, executable {6}",
+                            new Object[] { path, Files.isDirectory(path), Files.isRegularFile(path),
+                                    Files.isSymbolicLink(path), Files.isWritable(path), Files.isReadable(path),
+                                    Files.isExecutable(path) });
+                } else {
+                    LOGGER.log(Level.SEVERE, "Path {0} doesn't exists", path);
+                }
+
+                path = path.getParent();
+            } while (path != null && !exists);
+
             throw new DataStorageManagerException(err);
         }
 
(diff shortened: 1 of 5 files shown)


herddb-core/src/main/java/herddb/file/FileDataStorageManager.java
```diff
@@ -944,6 +944,25 @@ public void writeIndexPage(
             }
 
         } catch (IOException err) {
+            LOGGER.log(Level.SEVERE, "Failed to write on path: {0}", pageFile);
```
eolivelli: I am not sure we want to keep this debug.

Maybe you can create an helper method and enable it with a system property.
I am -0 on keeping it

diegosalvi: Actually it was a log added for debugging... then at commit time I thought to left it there (lines after line 947).

I'm not sure if let it there. It shouldn't happen but knowing that the parent path didn't exists helped to found the issue.

herddb-core/src/test/java/herddb/cluster/MultiServerTest.java
```diff
@@ -92,53 +92,85 @@ public void test_leader_online_log_available() throws Exception {
...
-            server_1.getManager().executeUpdate(new InsertStatement(TableSpace.DEFAULT, "t1", RecordSerializer.makeRecord(table, "c", 3)), StatementEvaluationContext.DEFAULT_EVALUATION_CONTEXT(), TransactionContext.NO_TRANSACTION);
-            server_1.getManager().executeUpdate(new InsertStatement(TableSpace.DEFAULT, "t1", RecordSerializer.makeRecord(table, "c", 4)), StatementEvaluationContext.DEFAULT_EVALUATION_CONTEXT(), TransactionContext.NO_TRANSACTION);
-
-            DMLStatementExecutionResult executeUpdateTransaction = server_1.getManager().executeUpdate(new InsertStatement(TableSpace.DEFAULT, "t1", RecordSerializer.makeRecord(table, "c", 5)), StatementEvaluationContext.DEFAULT_EVALUATION_CONTEXT(), TransactionContext.AUTOTRANSACTION_TRANSACTION);
-            server_1.getManager().executeStatement(new CommitTransactionStatement(TableSpace.DEFAULT, executeUpdateTransaction.transactionId), StatementEvaluationContext.DEFAULT_EVALUATION_CONTEXT(), TransactionContext.NO_TRANSACTION);
+            Table table = Table.builder().name("t1").column("c", ColumnTypes.INTEGER).column("s", ColumnTypes.INTEGER)
```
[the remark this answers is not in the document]
diegosalvi: It was an unwanted reformat while I was attempting to disable import blank line insertio by Eclipse.
Reverted the reformat.

PR #475, opened 2019-09-25
Enable Calcite ReduceExpressionsRule in order to simplify expressions

eolivelli:
With ReduceExpressionsRule Calcite tries to simplify more aggressively input expressions.
Stuff like: 
"SELECT from a where n = 1 and n is not null" becomes "SELECT from a where n = 1"
"SELECT from a where n < 1 and n < 2" becomes "SELECT from a where n < 1"

### DIFF
--- a/herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
+++ b/herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
@@ -133,6 +133,7 @@ import org.apache.calcite.rel.core.AggregateCall;
 import org.apache.calcite.rel.core.Sort;
 import org.apache.calcite.rel.core.TableModify;
 import org.apache.calcite.rel.logical.LogicalTableModify;
+import org.apache.calcite.rel.rules.ReduceExpressionsRule;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rel.type.RelDataTypeField;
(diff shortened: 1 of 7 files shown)


herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
```diff
@@ -505,6 +506,7 @@ private PlannerResult runPlanner(String defaultTableSpace, String query) throws
         RelDataType originalRowType = logicalPlan.getRowType();
         RelOptCluster cluster = logicalPlan.getCluster();
         final RelOptPlanner optPlanner = cluster.getPlanner();
+        optPlanner.addRule(ReduceExpressionsRule.FILTER_INSTANCE);
```
amitvc: @eolivelli what is the purpose of adding the Filter_Instance here?

eolivelli: This is the core of this change.
That Rule simplifies every expression, removing redundant filterd

amitvc: I see. I think this looks good to me.

eolivelli:
One test is failing.
I have to fix it before merging

herddb-core/src/test/java/herddb/core/SimpleOperatorsTest.java
```diff
@@ -294,28 +292,14 @@ public void simpleArithmeticOperationsTest() throws Exception {
             }
 
             // In expressions
-            // Warning: jsqlParser doesn't handle this kind of expressions in select clause
             try (DataScanner scan1 = scan(manager, "SELECT * FROM tblspace1.tsql WHERE '1' in (1,2,3)", Collections.emptyList())) {
```
eolivelli: @amitvc  this kind of query is not useful in real life.
We can change the behaviour

PR #477, opened 2019-09-26
Do not hang in case of internal error during system management tasks:

### LINKED ISSUE #476: Checkpoint stop running after IllegalMonitorStateException
**BUG REPORT**

We got a very slow startup on an HerdDB single istance (v0.12.0). Looking at the stacktrace we saw that it was replaying a lot of transaction from the WAL (~4.8GB of files inside the txlog).

Looking around we found out that the Checkpoint wasn't running since 7 days ago, during the last Checkpoint we saw this error:

```
19-09-19-23-42-45       herddb.core.TableSpaceManager Sep 19, 2019 11:42:45 PM herddb.core.TableSpaceManager checkpoint
INFO: gallina checkpoint finish herd started ad (400,163687), finished at (400,163687), total time 1552 ms

19-09-19-23-46-28       herddb.core.DBManager Sep 19, 2019 11:46:28 PM herddb.core.DBManager$Activator run
SEVERE: fatal activator erro
java.lang.IllegalMonitorStateException
        at java.base/java.util.concurrent.locks.StampedLock.unlockRead(StampedLock.java:708)
        at herddb.utils.LocalLockManager.releaseReadLock(LocalLockManager.java:146)
        at herddb.utils.LocalLockManager.releaseLock(LocalLockManager.java:158)
        at herddb.model.Transaction.releaseLocksOnTable(Transaction.java:211)
        at herddb.core.TableManager.onTransactionRollback(TableManager.java:1488)
        at herddb.core.TableSpaceManager.apply(TableSpaceManager.java:347)
        at herddb.core.TableSpaceManager.forceTransactionRollback(TableSpaceManager.java:1027)
        at herddb.core.TableSpaceManager.processAbandonedTransactions(TableSpaceManager.java:730)
        at herddb.core.DBManager.executeActivator(DBManager.java:1082)
        at herddb.core.DBManager.access$500(DBManager.java:115)
        at herddb.core.DBManager$Activator.run(DBManager.java:1005)
        at java.base/java.lang.Thread.run(Thread.java:834)
```

We need to handle this error case and, also, have a way to monitor if the Checkpoint don't run for too many time.

eolivelli:
Do not hang in case of internal error during system management tasks:
    - early exit in case of system shutdown and recovery in progress
    - wait for transactions to be in a stable status before forcing a rollback
    - do not exist core lifecycle management tasks

Fixes #476

### DIFF
--- a/herddb-core/src/main/java/herddb/core/TableSpaceManager.java
+++ b/herddb-core/src/main/java/herddb/core/TableSpaceManager.java
@@ -723,6 +723,7 @@ public class TableSpaceManager {
                         new Object[]{t.transactionId,
                                 new java.sql.Timestamp(t.localCreationTimestamp),
                                 new java.sql.Timestamp(t.lastActivityTs)});
+                validateTransactionBeforeTxCommand(t.transactionId);
                 long lockStamp = acquireReadLock("forceRollback" + t.transactionId);
                 try {
                     forceTransactionRollback(t.transactionId);
@@ -1770,7 +1771,6 @@ public class TableSpaceManager {
         if (tc == null) {
             throw new StatementExecutionException("no such transaction " + txId + " in tablespace " + tableSpaceName);
         }
-        int count = 0;
         while (tc.hasPendingActivities() && !closed) {
             LOGGER.log(Level.INFO, "Transaction {0} ({1}) has {2} pending activities",
                     new Object[]{txId, tableSpaceName, tc.getRefCount()});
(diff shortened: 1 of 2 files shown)


herddb-core/src/main/java/herddb/core/TableSpaceManager.java
```diff
@@ -1836,6 +1836,9 @@ public ApplyEntryOnRecovery() {
 
         @Override
         public void accept(LogSequenceNumber t, LogEntry u) {
+            if (dbmanager.isStopped()) {
+                throw new RuntimeException("System was requested to stop, aborting recovery at " + t);
```
diegosalvi: Can we ad a more "speaking" exception? Like ShutdownException extends HerdDBInternalException or something like that?

eolivelli: @diegosalvi 
It would make sense to have a specific Exception if downstream code can handle it, in this case the caller code is the CommitLog and it cannot recover in any way from the exception.

In which way a new exception here can add value ?

diegosalvi: Good point

PR #481, opened 2019-10-02
Update jSQLParser to 3.1

eolivelli:
Update jSQLParser to 3.0 and remove deprecated code of legacy pre-Calcite planner.

This is particularly needed in order to be compatible with latest versions of myBatis pagehelper plugin

Release notes:
https://github.com/JSQLParser/JSqlParser/releases/tag/jsqlparser-3.1

> @wumpz wumpz released this 16 days ago · 7 commits to master since this release
> 
> support for cross apply
> allowed keyword character as object name
> allowed keyword size as object name
> introduced FILTER expression for window functions
> allow more complex expressions for CASE.
> allowed start as object name as column name or table name
> introduced more positions for ! instead of NOT
> allowed more complex expressions within if function
> introduced multicolumn alias like select * from mytab as tab(c1, c2)


eolivelli:
It seems that jsqlparser does not support "RENAME TO" any more

https://github.com/JSQLParser/JSqlParser/issues/860

eolivelli:
WE have to move to 'ALTER TABLE tblspace1.tsql CHANGE k1 l2 int'

eolivelli:
@diegosalvi this patch is good to go

@amitvc FYI

eolivelli:
With this patch a downstream application that uses columns with name 'size' does work anymore,
I have reported an issue upstream to JSQLParser
https://github.com/JSQLParser/JSqlParser/issues/867

aluccaroni:
I see that we have a new release downstream, can we merge?

https://github.com/JSQLParser/JSqlParser/releases/tag/jsqlparser-3.1

eolivelli:
Yes.
In fact now Travis is fine

cc @diegosalvi  @amitvc

eolivelli:
I feel we should add a testcase about the problem we found in EmailSuccess
```
With this patch a downstream application that uses columns with name 'size' does work anymore,
I have reported an issue upstream to JSQLParser
```

PR #484, opened 2019-10-15
Fix race condition on creating a tablespace with a very slow leader

eolivelli:
Give a little more time to the new leader to write its state right after the creation of the tablespace.
Clean up logs during the boot, some "SEVERE" logs should be reported as INFO


eolivelli:
@diegosalvi we should cherry pick to 0.12 branch.

I did not add a specific test

PR #488, opened 2019-10-21
Make automatic failover more stable

eolivelli:
This patch simply make it more clear that a node should take into account the last ping for the leader only of counting on the replica state of the current known leader.
This makes the system converge more quickly to a new leader

### DIFF
--- a/herddb-core/src/main/java/herddb/core/DBManager.java
+++ b/herddb-core/src/main/java/herddb/core/DBManager.java
@@ -909,7 +909,7 @@ public class DBManager implements AutoCloseable, MetadataChangeListener {
     }
 
     private void tryBecomeLeaderFor(TableSpace tableSpace) throws DDLException, MetadataStorageManagerException {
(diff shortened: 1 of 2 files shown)


herddb-core/src/main/java/herddb/core/DBManager.java
```diff
@@ -1186,18 +1188,17 @@ private boolean manageTableSpaces() {
...
                             && tableSpaceInfo.maxLeaderInactivityTime > 0
                             && !tableSpaceManager.isFailed()) {
                         List<TableSpaceReplicaState> allReplicas =
                                 metadataStorageManager.getTableSpaceReplicaState(tableSpaceUuid);
                         TableSpaceReplicaState leaderState = allReplicas
                                 .stream()
-                                .filter(t -> t.mode == TableSpaceReplicaState.MODE_LEADER
-                                        && !t.nodeId.equals(nodeId)
-                                )
+                                .filter(t -> t.nodeId.equals(tableSpaceInfo.leaderId))
```
eolivelli: This is the core of the change.
In case of multiple servers trying to become leader the prev check could return values from the dead leader resulting in lots of leadership changes.

Issue #468, opened 2019-09-20
PRIMARY KEY not parsed correctly when executed via CLI - Was "Discrepancy between CREATE TABLE and SHOW CREATE TABLE commands when using auto_increment"


aluccaroni:
**BUG REPORT**

1. Please describe the issue you observed:

I seem to be unable to create a table with an autoincrement column.
If I run this query:
`create table mytablespace.test2 (ID long auto_increment, LastName varchar(255), FirstName varchar(255), Address varchar(255), City varchar(255),PRIMARY KEY (id) )`

And then run a "SHOW CREATE TABLE test2", the output is:
`CREATE TABLE mytablespace.test2(id long,lastname string,firstname string,address string,city string,PRIMARY KEY(id)) `

The column "id" seems to have no auto_increment flag, and I can confirm that by running:
`insert into test2 values (null,'Luccaroni','Alessandro','via Glicini 8','Faenza')`

```
Executing query:insert into test2 values (null,'Luccaroni','Alessandro','via Glicini 8','Faenza')
ERROR:java.sql.SQLException: herddb.model.StatementExecutionException: From line 1, column 19 to line 1, column 81: Column 'id' has no default value and does not allow NULLs
```

If, instead, we run this create table (without declaring the PK):
`create table mytablespace.testpk (ID long , LastName varchar(255), FirstName varchar(255), Address varchar(255), City varchar(255) )`

And then run a "SHOW CREATE TABLE testpk", the output is:
`CREATE TABLE mytablespace.testpk(id long,lastname string,firstname string,address string,city string,_pk long auto_increment,PRIMARY KEY(_pk))`

The pk column _pk is created with the auto_increment flag, and it works:
`insert into testpk values (1,'Luccaroni','Alessandro','via Glicini 8','Faenza',null)
`
```
select * from testpk
| id | lastname  | firstname  | address       | city   | _pk |
| -- | --------  | ---------  | -------       | ----   | --- |
| 1  | Luccaroni | Alessandro | via Glicini 8 | Faenza | 1   |
```

aluccaroni:
It seems to be a problem with the query parsing, if I run the following query it works:
`create table mytablespace.test5 (ID long AUTO_INCREMENT PRIMARY KEY, LastName varchar(255), FirstName varchar(255), Address varchar(255), City varchar(255))`

Please note that the first query that I've used (and don't work) is the one outputted by SHOW CREATE TABLE:
[...]

PR #495, opened 2019-11-01
PR for https://github.com/diennea/herddb/issues/468

amitvc:
The fix for this PR deals with using lower case to search for columns that are auto_increment are also part of primary key. We lowercase the primary keys column names and were searching the list for it with uppercase name. Hence we would not find it. The fix deals with this issue.


eolivelli:
@aluccaroni  good catch !
I would prefer to commit this first patch and then we can follow up with a new issue

aluccaroni:
See https://github.com/diennea/herddb/issues/497

PR #498, opened 2019-11-04
Prevent multiple processing of the same record during concurrent DML operations

eolivelli:
After the great work by @diegosalvi in #500 I have rebased this branch, now it contains only:
- drop legacy LockManager
- clean up logs (there were some SEVERE logs)


herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -3158,6 +3152,9 @@ public void accept(Entry<Bytes, Long> entry) throws DataStorageManagerException,
                         transaction.touch();
                     }
                     Bytes key = entry.getKey();
+                    if (consumedKeys.contains(key)) {
```
eolivelli: this is actually the fix of the bug

eolivelli:
After the great work by @diegosalvi in #500 I have rebased this branch, now it contains only:
- drop legacy LockManager
- clean up logs (there were some SEVERE logs)

Issue #491, opened 2019-10-23
Handle gracefully Calcite exception inside the CLI


diegosalvi:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.
_When executing "broken" queries on CLI we get big stacktraces with not much infos to non developers. We should have better use informations and messages at least on CLI and leave stacktraces to logs (if any... they could go to debug)_
2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?
_nice-to-have_
3. Provide any additional detail on your proposed use case for this feature.
```
ERROR:java.sql.SQLException: herddb.model.StatementExecutionException: Non-query expression encountered in illegal context
        at herddb.sql.CalcitePlanner.translate(CalcitePlanner.java:342)
        at herddb.server.ServerSideConnectionPeer.handleOpenScanner(ServerSideConnectionPeer.java:445)
        at herddb.server.ServerSideConnectionPeer.requestReceived(ServerSideConnectionPeer.java:181)
        at herddb.network.netty.NettyChannel.lambda$handlePduRequest$0(NettyChannel.java:96)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.base/java.lang.Thread.run(Thread.java:834)
```

aluccaroni:
Some other errors examples:

```
herd: CREATE TABLESPACE fagiano¶½
Executing query:CREATE TABLESPACE fagiano¶½
ERROR:java.sql.SQLException: herddb.model.StatementExecutionException: unable to parse query EXECUTE createtablespace fagiano¶½
        at herddb.sql.DDLSQLPlanner.parseStatement(DDLSQLPlanner.java:237)
        at herddb.sql.DDLSQLPlanner.translate(DDLSQLPlanner.java:220)
        at herddb.sql.CalcitePlanner.translate(CalcitePlanner.java:244)
[...]

PR #515, opened 2019-11-27
Hide stacktraces from CLI by default

eolivelli:
Hide stacktraces by default in the CLI.
In order to see stacktraces the user must use "-v" (verbose) option


eolivelli:
This is a fix for #491 

We should add it in 0.13.1

eolivelli:
I have to fix checkstyle

PR #524, opened 2019-12-09
Add new configFile JDBC url string parameter to ease setup of embedded server

eolivelli:
Setting up an embedded server using pure JDBC URL is very difficult.
We can let the user create a simple properties file and link to it

jdbc:herddb:server:localhost:7000?configFile=PATH/TO/CONFIGFILE


eolivelli:
@amitvc  @diegosalvi PTAL

PR #528, opened 2019-12-16
Allow to configure the replica count from the configuration file

eolivelli:
- add a new configuration parameter tablespace.default.replica.count=1 that is the default number of replicas
- enhance the default server.properties file

### DIFF
--- a/herddb-core/src/main/java/herddb/sql/DDLSQLPlanner.java
+++ b/herddb-core/src/main/java/herddb/sql/DDLSQLPlanner.java
@@ -654,9 +654,7 @@ public class DDLSQLPlanner implements AbstractSQLPlanner {
                 Object tableSpaceName = resolveValue(execute.getExprList().getExpressions().get(0), true);
                 String leader = null;
                 Set<String> replica = new HashSet<>();
-                int expectedreplicacount = this.manager.getServerConfiguration() != null
-                        ? this.manager.getServerConfiguration().getInt(ServerConfiguration.PROPERTY_DEFAULT_REPLICA_COUNT, ServerConfiguration.PROPERTY_DEFAULT_REPLICA_COUNT_DEFAULT)
-                        : 1;
+                int expectedreplicacount = this.manager.getServerConfiguration().getInt(ServerConfiguration.PROPERTY_DEFAULT_REPLICA_COUNT, ServerConfiguration.PROPERTY_DEFAULT_REPLICA_COUNT_DEFAULT);
                 long maxleaderinactivitytime = 0;
                 int wait = 0;
                 for (int i = 1; i < execute.getExprList().getExpressions().size(); i++) {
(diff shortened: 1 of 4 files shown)


herddb-core/src/main/java/herddb/sql/DDLSQLPlanner.java
```diff
@@ -653,7 +654,9 @@ private Statement buildExecuteStatement(String defaultTableSpace, Execute execut
                 Object tableSpaceName = resolveValue(execute.getExprList().getExpressions().get(0), true);
                 String leader = null;
                 Set<String> replica = new HashSet<>();
-                int expectedreplicacount = 1;
+                int expectedreplicacount = this.manager.getServerConfiguration() != null
```
amitvc: Should we abstract this logic this.manager.getServerConfiguration() ! = null into some class or method and have a utility method to which we just pass property key 'PROPERTY_DEFAULT_REPLICA_COUNT' (in this case). We can reuse this functionality someplace else so we don't duplicate the code.

herddb-services/src/main/resources/conf/server.properties
```diff
@@ -61,12 +61,21 @@ server.ssl=false
...
 
-# for server.mode=cluster
+# for server.mode=cluster you have to set a connection string to your ZooKeeper cluster
+# you have to write here the list of all of the zookeeper servers, not only one
 server.zookeeper.address=localhost:2181
 server.zookeeper.session.timeout=40000
 server.zookeeper.path=/herd
 
+# data replication settings
+# this is the default value for 'expectedreplicacount' for CREATE TABLESPACE command
```
amitvc: Based on the comment here it seems that the replica count can and should be less than or equal to the number of nodes in the cluster. Are we checking for that somewhere or throw an error if the user accidentally chooses a number that is higher?

herddb-core/src/main/java/herddb/sql/DDLSQLPlanner.java
eolivelli: It is a good suggestion.
I will change this line.
But I prefer not to change other lines in the scope of this patch.

I will double check, maybe ServerConfiguration is not expected to be null

herddb-services/src/main/resources/conf/server.properties
eolivelli: This is somehow tricky to understand.
We are setting a default value for expectedreplicacount that is a "target replication" factor for data.
In order to create a tablespace all you need is only a *leader* node.
When you set *expectedreplicacount* the system will try to have that number of replicas *eventually*.
It is a limit that it is not enforced.

We have to parameters for replication:
-bookkeeper write replication (writequorumsize)
- data replication

herddb-core/src/main/java/herddb/sql/DDLSQLPlanner.java
eolivelli: I have double checked and ServerConfiguration is "final" in DBManager so it is safe to consider it not null.
I have simplified this line.

Good catch @amitvc !

eolivelli:
@amitvc I have addressed your comment, please take a look
Thank you for your review

herddb-services/src/main/resources/conf/server.properties
amitvc: I see. Thanks for explaining this.

Issue #529, opened 2019-12-24
TableSpace recovery : Corrupted transaction, syncing on a position smaller ((1504,17371917)) than transaction last sequence number ((1504,17371917))


hamadodene:
Is this a question, a feature request or a bug report?

- Is a bug report

**BUG REPORT**

1. Please describe the issue you observed:
During recovery of a tablespace i get this error: 
```
19-12-23-15-55-47       herddb.core.DBManager Dec 23, 2019 3:55:47 PM herddb.core.DBManager manageTableSpaces
SEVERE: cannot handle tablespace q1
herddb.log.LogNotAvailableException: java.lang.IllegalStateException: Corrupted transaction, syncing on a position smaller ((1504,17371917)) than transaction last sequence number ((1504,17371917))
        at herddb.cluster.BookkeeperCommitLog.recovery(BookkeeperCommitLog.java:502)
        at herddb.core.TableSpaceManager.recover(TableSpaceManager.java:294)
        at herddb.core.TableSpaceManager.start(TableSpaceManager.java:231)
        at herddb.core.DBManager.handleTableSpace(DBManager.java:537)
        at herddb.core.DBManager.manageTableSpaces(DBManager.java:1120)
        at herddb.core.DBManager.executeActivator(DBManager.java:1066)
        at herddb.core.DBManager.access$500(DBManager.java:113)
        at herddb.core.DBManager$Activator.run(DBManager.java:1009)
        at java.base/java.lang.Thread.run(Thread.java:830)
Caused by: java.lang.IllegalStateException: Corrupted transaction, syncing on a position smaller ((1504,17371917)) than transaction last sequence number ((1504,17371917))
        at herddb.model.Transaction.sync(Transaction.java:567)
        at herddb.core.TableSpaceManager.apply(TableSpaceManager.java:359)
        at herddb.core.TableSpaceManager$ApplyEntryOnRecovery.accept(TableSpaceManager.java:1843)
        at herddb.core.TableSpaceManager$ApplyEntryOnRecovery.accept(TableSpaceManager.java:1832)
        at herddb.cluster.BookkeeperCommitLog.recovery(BookkeeperCommitLog.java:470)
        ... 8 more
```

- What did you do?
It happened during a checkpoint.
From my observations:
The checkpint started in position (1504,17371916) and ended in position (1504,17372323).
During the checkpoint activity, however, there was an already started transaction that continued to go even during the checkpoint.
The transaction finished in position (1504,17371917).

- What did you expect to see?
[...]

PR #530, opened 2019-12-24
Ensure correct LastSequenceNumber is published on checkpoint files on slow machines

eolivelli:
This patches fixes an error condition in case of slow system.
When we are doing the checkpoint we sample the LastSequenceNumber for BookKeeperCommitLog.
It can happen that we are observing a stale value.
This patches ensures that when an entry is visible as committed that the LogSequenceNumber is correctly seen by the TableSpaceManager.

I managed to reproduce an error with a test case but it needs to instrument BookKeeperCommitLog inside the code that I have changed (replace *theAccept* with *theApply*)
Just by adding a Thread.sleep(2000) inside the body of "theAccept" lambda function we can observe failures

### DIFF
--- a/herddb-core/src/main/java/herddb/core/TableSpaceManager.java
+++ b/herddb-core/src/main/java/herddb/core/TableSpaceManager.java
@@ -267,7 +267,7 @@ public class TableSpaceManager {
         }
         dataStorageManager.loadTransactions(logSequenceNumber, tableSpaceUUID, t -> {
             transactions.put(t.transactionId, t);
-            LOGGER.log(Level.FINER, "{0} {1} tx {2} at boot", new Object[]{nodeId, tableSpaceName, t.transactionId});
+            LOGGER.log(Level.FINER, "{0} {1} tx {2} at boot lsn {3}", new Object[]{nodeId, tableSpaceName, t.transactionId, t.lastSequenceNumber});
             try {
                 if (t.newTables != null) {
                     for (Table table : t.newTables.values()) {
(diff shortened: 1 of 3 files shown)


herddb-core/src/main/java/herddb/core/TableSpaceManager.java
```diff
@@ -1632,7 +1632,14 @@ TableSpaceCheckpoint checkpoint(boolean full, boolean pin, boolean alreadLocked)
                     throw new DataStorageManagerException("actualLogSequenceNumber cannot be null");
                 }
                 // TODO: transactions checkpoint is not atomic
-                actions.addAll(dataStorageManager.writeTransactionsAtCheckpoint(tableSpaceUUID, logSequenceNumber, new ArrayList<>(transactions.values())));
+                Collection<Transaction> currentTransactions = new ArrayList<>(transactions.values());
+                for (Transaction t : currentTransactions) {
+                    LogSequenceNumber txLsn = t.lastSequenceNumber;
```
amitvc: Can there be a case when LogSequenceNumber is null? If yes why and should we log info or error to indicate that?

eolivelli: Yes
When you create a transaction it does not have a lsn because we can assign a value only for writes to the log that have been persisted on the CommitLog.
While dealing with transactions we only require that the commit() operation is fsynced to disk, this is particularly needed for BK commit log.
So it is natural for instance that a transaction just created without any dml execution has a null lsn.
So logging this case may lead to a lot of curn into the logs

https://github.com/diennea/herddb/blob/e29eab66acccc9ecb6dffe6b31c8f7fa6fc21e41/herddb-core/src/main/java/herddb/core/TableSpaceManager.java#L326

amitvc: Makes sense.

hamadodene:
Do this PR resolve #529 ?

diegosalvi:
Good catch!

hamadodene:
I performed a benchmark with this integration in ssd, and the result is ok.
Good catch! @eolivelli

PR #531, opened 2019-12-24
Remove duplicates while sending dump to followers in case of slow system

eolivelli:
- do not assume that a full checkpoint has been done and the phisical page set contains only non dirty data
- do not force a full checkpoint during followers boot and during backups
- clean up FullTableScanConsumer, leaving space for further optimization (batch processing) in the future
- introduce new configuration server.boot.force.download.snapshot to force the download of a fresh new snapshot from the leader (useful for tests)

### DIFF
--- a/herddb-core/src/main/java/herddb/core/SingleTableDumper.java
+++ b/herddb-core/src/main/java/herddb/core/SingleTableDumper.java
@@ -21,7 +21,6 @@
 package herddb.core;
 
 import herddb.core.stats.TableManagerStats;
-import herddb.index.KeyToPageIndex;
 import herddb.model.Index;
 import herddb.model.Record;
 import herddb.model.Table;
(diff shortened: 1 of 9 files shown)


herddb-core/src/main/java/herddb/core/SingleTableDumper.java
```diff
@@ -82,25 +86,25 @@ public void acceptTableStatus(TableStatus tableStatus) {
...
+    public void acceptPage(long pageId, List<Record> records) {
         try {
-            batch.add(new KeyValue(record.key, record.value));
-            if (batch.size() == fetchSize) {
-                sendBatch();
+            for (Record record : records) {
+                // since 0.14 we are no more compacting the table before sending a dump
+                // we have to check that the record is really on the current page
+                // before sending it to the client/follower
+                Long currentPageForRecord = keyToPageIndex.get(record.key);
```
eolivelli: This is actually the fix.
It is expensive but it is saving the date :-)

eolivelli:
I am not sure this is the right approach.

We have to keep the write checkpoint lock and prevent all writes.

Alternatively we can keep the full checkpoint but perform the table scan using the right set of active pages (this should be the current approach but evidently there is a bug)

eolivelli:
I have removed the keyToPage lookup, but still we have a problem: the "fullcheckpoint" sometimes skips compacting smallpages and it skips other cases.
If we want to perform the dump this way the checkpoint MUST remove all duplicates and remove all dead records

PR #532, opened 2020-01-07
Wait for remove page from PageReplacementPolicy during compaction

diegosalvi:
During page compaction we remove the compated page from page replacement policy before handling it. After handling page records we have to check again if the page is in loaded pages and if it has been reloaded again for some reasons.

The check has to stay in place as is but we should move the removal of page knowledge only after record handling to avoid multiple reload of the same page.

Added missing unload page counter too.


diegosalvi:
This patch should not be merged now, we are in line to deliver the next release.
The case itself is already covered by test suites, this just attempt to avoid mutiple page loads (that still can happen between policy page removal and real page removal thus the check is still in place)

eolivelli:
Ok waiting for travis

diegosalvi:
We could think about integrate this

eolivelli:
Sure.
Can you please rebase?
This way we will check again with Travis

Issue #540, opened 2020-01-22
 Corrupted transaction log during follower recovery: key is already present in table 


hamadodene:
During Herd's normal activity, i had problems while recovering tablespaces.
When node x tries to download data from node y, a duplicate key arrives at node x.
This means that the y node  (we will call it "myhostname/myip") during the full checkpoint does not actually remove all duplicates.

Stacktrace below:
```
20-01-22-15-56-12       herddb.network.netty.NettyChannel Jan 22, 2020 3:56:12 PM herddb.network.netty.NettyChannel lambda$handlePduRequest$0
SEVERE: NettyChannel{name=myhostname:7000, id=13, socket=[id: 0xe6347342, L:/10.162.1.11:38666 - R:myhostname/myip:7000] pending 0 msgs}: error java.lang.IllegalStateException: corrupted transaction log: key 0800000000005c4bd70400000001124c4f43414c2042494e442041444452455353 is already present in table q2_additional_info
java.lang.IllegalStateException: corrupted transaction log: key 0800000000005c4bd70400000001124c4f43414c2042494e442041444452455353 is already present in table q2_additional_info
        at herddb.core.TableManager.applyInsert(TableManager.java:1957)
        at herddb.core.TableManager.writeFromDump(TableManager.java:1872)
        at herddb.core.ReplicaFullTableDataDumpReceiver.receiveTableDataChunk(ReplicaFullTableDataDumpReceiver.java:105)
        at herddb.client.RoutedClientSideConnection.requestReceived(RoutedClientSideConnection.java:203)
        at herddb.network.netty.NettyChannel.lambda$handlePduRequest$0(NettyChannel.java:96)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.base/java.lang.Thread.run(Thread.java:830)

```

eolivelli:
@hamadodene using the database you sampled fro the environment I am able to reproduce exactly the error on the same key locally on a test case.
[...]

Issue #539, opened 2020-01-22
Leadership election: cannot download data of tableSpace buckets from myself


hamadodene:
While using the cluster installation, I had problems with the leadership election.
A cluster node remained isolated  caused by network down (we will call it n1).
The  node m1 was the leader of the tablespace buckets, and when it went "down" the leadership moved to another node.
However, this node did not have enough data to start as a leader (we will call it n2).
We will call the third node n3.
Restarting node n2, n3 the leadership changed and passed on node n3 and everything started to work again.
This means that there is a problem in choosing the leader.
I expect that a node that does not have enough data will not be named. Maybe it's probably not important that it becomes the node with the most data, but that in a loop you try to download the data from another node and not from itself.

```
20-01-21-18-11-54       herddb.core.DBManager Jan 21, 2020 6:11:54 PM herddb.core.DBManager manageTableSpaces
SEVERE: cannot handle tablespace buckets
herddb.storage.DataStorageManagerException: cannot download data of tableSpace buckets from myself
        at herddb.core.TableSpaceManager.downloadTableSpaceData(TableSpaceManager.java:589)
        at herddb.core.TableSpaceManager.recover(TableSpaceManager.java:302)
        at herddb.core.TableSpaceManager.start(TableSpaceManager.java:231)
        at herddb.core.DBManager.handleTableSpace(DBManager.java:537)
        at herddb.core.DBManager.manageTableSpaces(DBManager.java:1120)
        at herddb.core.DBManager.executeActivator(DBManager.java:1066)
        at herddb.core.DBManager.access$500(DBManager.java:113)
        at herddb.core.DBManager$Activator.run(DBManager.java:1009)
        at java.base/java.lang.Thread.run(Thread.java:830)
 
```

aluccaroni:
If we know which node has the most data it's a good idea that that node will become master, but this maybe add un-needed complexity.

I prefer a simpler approach where a node at random become master and this node will guarantee, checking each replica, to become up-to-date before actually start up the tablespace...usually all the replica are very close (and, anyway, every replica needs to become consistent eventually) so this shouldn't be more expensive in the end.

eolivelli:
We are adding a new replica automatically depending on "expectedreplicacount", but this happens only in case of an explicit lack of nodes in the "replica" field.
We are not detecting unavailable nodes and remove them from the "replica" set.

@hamadodene please check your logs for 
[...]

PR #542, opened 2020-01-24
Fix faulty cases of follower startup

eolivelli:
This is a series of fixes about the case in which a follower downloads a big tablespace and it is interrupted during the first recovery.
The correct behaviour should be to erase all data from disk and start from scratch.

We are also adding a facility for hard-coded instrumentation (I tried with Powermock but in these particular cases I preferred to use static instrumentation)


eolivelli:
@diegosalvi I feel this is the cause of #539  and #540  in @hamadodene long running tests and manual fault injection

PR #544, opened 2020-01-27
Protect DBManager.executeActivator from processAbandonedTransactions failures

diegosalvi:
This patch fixes two issues:
* if one abandoned transaction check fails every other abandoned transaction in every other tablespace is not checked (and every "job" after abandoned transaction check but currently it is the last): added a fix to not propagate the exception but simply log it and continue to the next transaction;
* if a transaction is in a pending unrecoverable state the whole activator thread hang forever waiting for it: added a fix to not wait for pending transaction to complete on activator but simply continue to the next transaction.


eolivelli:
Travis failed again on ScanHugeTableTest, this is very weird.

PR #545, opened 2020-01-30
Handle BK/ZK errors during automatic transaction rollback in clients.

eolivelli:
In case of BK/ZK error the client is not able to commit a transaction
it blindly issues a rollback command.
This command may go into a new ledger in case BookKeeperCommitLog
does not closes itself in time.

The fix is to automatically in-memory rollback transactions with a failed
write on the commit command.
This prevents any further command in the scope of the transaction
and expectially rollbacks to be validated.

The problem may occour even on FileCommitLog (standalone mode).

We are also broadening the handling of BK failures, in the case
of the issue we had a ZKException.

### DIFF
--- a/herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
+++ b/herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
@@ -500,13 +500,11 @@ public class BookkeeperCommitLog extends CommitLog {
                                     localEntryCount++;
                                 }
                             }
-                            LOGGER.log(Level.FINER, tableSpaceDescription()+" read " + localEntryCount + " entries from ledger " + ledgerId
+                            LOGGER.log(Level.FINER, tableSpaceDescription() + " read " + localEntryCount + " entries from ledger " + ledgerId
                                     + ", expected " + entriesToRead);
 
-                            LOGGER.log(Level.FINER, tableSpaceDescription()+" finished waiting for " + entriesToRead
-                                    + " entries to be read from ledger " + ledgerId);
                             if (localEntryCount != entriesToRead) {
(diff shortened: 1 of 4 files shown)


eolivelli:
This patch also cleans up logs and add useful information for debugging

diegosalvi:
Checkstyle fails

diegosalvi:
From my understanding no rollback real will be issued on log if commit write (and previous related events) fails but I don't understand what will happen if open transaction log was successfully issued but other events failed to write on BK. Do a real rollback will be written?
Can this case be covered with a test?

eolivelli:
> From my understanding no rollback real will be issued on log if commit write (and previous related events) fails but I don't understand what will happen if open transaction log was successfully issued but other events failed to write on BK. Do a real rollback will be written?
> Can this case be covered with a test?

Do you mean this scenario ?
- client sends **begin transaction** -> failure on log
- async roll ledger
- client sends **insert record** -> no failure, entry is stored on the new ledger
- restart server
- boom ! **insert record** is about a transaction that never began

This case is not covered by a test and it is not fixed by this patch.
We are only covering the fact that an error on "commit" prevents any other commands (rollbacks in particular).
Maybe it is a good story to explore and to try to reproduce with a test case and fix it

eolivelli:
test is hanging...
I have to debug
herddb.cluster.BookkeeperFailuresTest.testBookieNotAvailableDuringTransactionWithRollback

diegosalvi:
I was more something like:
1. Begin transaction -> log good
2. Insert record -> log good
3. Insert record -> failure on log
4. Commit -> failure on log

If 4 cannot roll over the log and attempt another ledger to write the rollback we are left with a pending transaction with no commit nor rollback

eolivelli:
With this fix the transaction is discarded in memory.

In case of restart the pending transaction is forced to be rolled back
We should already have tests

PR #546, opened 2020-01-31
Adding new developer to pom

amitvc:
Added myself to developer section in pom.xml

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,11 +78,10 @@
         <developer>
             <id>amitvc</id>
             <name>Amit Chavan</name>
-            <organization>Diennea</organization>
         </developer>
     </developers>
     <issueManagement>
-        <url>https://dev.herddb.org/jira</url>
+        <url>https://github.com/diennea/herddb/issues</url>
     </issueManagement>
     <properties>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -75,6 +75,11 @@
             <name>Francesco Caliumi</name>
             <organization>Diennea</organization>
         </developer>
+        <developer>
+            <id>amitvc</id>
+            <name>Amit Chavan</name>
+            <organization>Diennea</organization>
```
eolivelli: I think this is not a good value, as you are not an "employee" of Diennea.

This field is optional, you can drop the <organization> tag

pom.xml
```diff
@@ -75,6 +75,11 @@
...
+        </developer>
     </developers>
     <issueManagement>
         <url>https://dev.herddb.org/jira</url>
```
eolivelli: can you also please fix this value ?
we moved our issue tracked to github

pom.xml
amitvc: Ok will do.

pom.xml
amitvc: ok

eolivelli:
@amitvc feel free to push the "merge button" as soon as CI is green

PR #549, opened 2020-02-05
Wait for FollowerThread to exit and add more logs

eolivelli:
- Ensure that FollowerThread is really stopped because spawning a new instance of TableSpaceManager
- Add more logs about metadata changes
- Add more logs about general lifecycle

### DIFF
--- a/herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
+++ b/herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
@@ -299,7 +299,7 @@ public class BookkeeperCommitLog extends CommitLog {
             res = new CompletableFuture();
             res.completeExceptionally(
                     new LogNotAvailableException(new Exception("this commitlog has been closed for tablespace "
-                            + tableSpaceDescription()))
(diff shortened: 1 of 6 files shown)


herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
```diff
@@ -724,6 +731,11 @@ void ensureOpenReader(LogSequenceNumber currentPosition) throws org.apache.bookk
             }
             currentLedger = bookKeeper.openLedgerNoRecovery(ledgerToTail,
                     BookKeeper.DigestType.CRC32C, SHARED_SECRET.getBytes(StandardCharsets.UTF_8));
+            
+            if (LOGGER.isLoggable(Level.FINE)) {
+                String ledgerLeader = extractLeaderFromMetadata(currentLedger.getLedgerMetadata().getCustomMetadata());
```
diegosalvi: Is this an heavy operation? We don't spam ledger opening.. we could log it even at info level

eolivelli: changed to INFO

PR #551, opened 2020-02-06
Rework BookKeeperCommitLog and introduce server.bookkeeper.ledgers.max.size

eolivelli:
- introduce server.bookkeeper.ledgers.max.size
 - rework BookKeeper failure handler
 - add direct unit tests over BookKeeperCommitLog
 - ensure that every write has been completed before rolling a new ledger
- clean up ZKTestEnv in order to start a more lightweight bookie

### DIFF
--- a/herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
+++ b/herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
@@ -341,16 +341,16 @@ public class BookkeeperCommitLog extends CommitLog {
         try {
             _writer = getValidWriter();
         } catch (LogNotAvailableException errorWhileRollingLedger) {
-            LOGGER.log(Level.SEVERE,"Cannot get a valid writer for "+tableSpaceDescription(), errorWhileRollingLedger);
+            LOGGER.log(Level.SEVERE, "Cannot get a valid writer for " + tableSpaceDescription(), errorWhileRollingLedger);
         }
         if (failed) {
-            res = FutureUtils.exception( new LogNotAvailableException(new Exception("this commitlog is failed, tablespace "
-                            + tableSpaceDescription() + ", node " + this.localNodeId))
-                            .fillInStackTrace());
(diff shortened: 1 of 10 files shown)


eolivelli:
This is only a draft, tests are missing

herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
```diff
@@ -149,51 +163,76 @@ public long getLedgerId() {
...
+        public WriteHandle getOut() {
             return out;
         }
 
         private void writeNoop() {
             // write a dummy entry, this will force LastAddConfirmed to be piggybacked
             try {
                 log(LogEntryFactory.noop(), false);
-            } catch (Throwable t) {
+            } catch (LogNotAvailableException t) {
```
diegosalvi: +100!

herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
```diff
@@ -345,19 +402,20 @@ private void handleBookKeeperAsyncFailure(Throwable cause, LogEntry edit) {
                     + tableSpaceDescription(), cause);
         } else if (cause instanceof BKException.BKLedgerFencedException) {
             LOGGER.log(Level.SEVERE, "this server was fenced for tablespace " + tableSpaceDescription() + " !", cause);
-            close();
```
diegosalvi: It doesn't need close anymore here?

eolivelli: now this code is executed during the write callback.

"closing" a WriteHandle is a metadata operation, and it errors out every other pending reads.

unfortunately it is named "close" but it should be named "seal", it is not about releasing resources but about writing to ZK current state of the ledger (LastAddConfirmed)

This is not supposed to be done inside the write callback.
We are marking the writer as "failed" any other "recovery" will be performed at a later stage
when the leader starts again and it opens the ledger for "recovery"

PR #554, opened 2020-02-12
Better log during recovery

eolivelli:
Catch any RuntimeException and log it in BookKeeperCommitLog logger, this way we can have the error in BookKeeperCommitLog log files, this allows better debugging in case of error in production.


eolivelli:
@amitvc can you commit this patch? Using the 'merge button' or merging manually to master and then pushing

amitvc:
Yes will do.

PR #557, opened 2020-02-19
Check consistency at every recovery

eolivelli:
New configuration option *server.localconsistencycheck.enabled* that forces a full table and index scan on every table after every "recovery" operation for each tablespace.

Implement a new SecondaryIndexFullScan that is needed for a full scan of every secondary index.

Scanning a table loads every page into memory in a temporary buffer, and it verifies that the page pointed by the keyToPage index is valid and contains the record as expected.

### DIFF
--- a/herddb-core/src/main/java/herddb/server/ServerConfiguration.java
+++ b/herddb-core/src/main/java/herddb/server/ServerConfiguration.java
@@ -137,7 +137,7 @@ public final class ServerConfiguration {
     public static final String PROPERTY_BOOKKEEPER_ACKQUORUMSIZE = "server.bookkeeper.ack.quorum.size";
     public static final int PROPERTY_BOOKKEEPER_ACKQUORUMSIZE_DEFAULT = 1;
     public static final String PROPERTY_BOOKKEEPER_LEDGERS_RETENTION_PERIOD = "server.bookkeeper.ledgers.retention.period";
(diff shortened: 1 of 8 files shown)


eolivelli:
It is enabled by default, this is for development purpose only, we have to disable it by default before 0.15.0 release.

eolivelli:
Tests are failing

eolivelli:
@hamadodene I have rebased onto latest master

herddb-core/src/main/java/herddb/server/ServerConfiguration.java
```diff
@@ -151,6 +151,9 @@
     public static final String PROPERTY_CHECKPOINT_PERIOD = "server.checkpoint.period";
     public static final long PROPERTY_CHECKPOINT_PERIOD_DEFAULT = 1000L * 60 * 15;
 
+    public static final String PROPERTY_ENABLE_CONSISTENCY_CHECK = "server.localconsistencycheck.enabled";
```
diegosalvi: This should be added on the example configuration file too

eolivelli: We will drop this feature as soon as we have the distributed consistency checker.
I think we should not advertise it too much

eolivelli:
I am not sure we need this patch at all...

herddb-core/src/main/java/herddb/server/ServerConfiguration.java
diegosalvi: Good point

eolivelli:
Closing this PR for now.
There is no need for it

PR #559, opened 2020-02-19
Ensure memory is released on role change

eolivelli:
When a node switches mode, for instance from follower to leader, it stops the internal **TableSpaceManager** and forgets about it.
Then we start a new TableSpaceManager instance that works on the same data files as in his previous life.
The **PagePlacementPolicy** is still holding a reference to the old  **TableManagers** and **IndexManagers**, so in case of page unload it is possible that the 'old life' of a **TableManager** flushes to disk a very old version of a mutable datapage, or writes to disk onto a filename that is in use by the "current" TableManager.
_This scenario leads to unpredictable data corruption._

This change ensures that every data is flushed from the PagePlacementPolicy when a TableManager is closed.

The attached new test case is failing with this fix

### DIFF
--- a/herddb-core/src/main/java/herddb/core/TableManager.java
+++ b/herddb-core/src/main/java/herddb/core/TableManager.java
@@ -1989,7 +1989,15 @@ public final class TableManager implements AbstractTableManager, Page.Owner {
 
     @Override
     public void close() {
+
+        // unload all pages
+        final List<DataPage> unload = pages.values().stream()
+                .collect(Collectors.toList());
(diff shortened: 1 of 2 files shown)


diegosalvi:
This could really really explain possible page overwrite!
Next page id is "resetted" then the page id could be generated twice!
During normal operation it shouldn't be a problem, newly create pages will overwrite old broken page but during checkpoint are create some pages that will not loaded to policy memory... an unload of old "broken" page after such pages flush would overwrite data in a totally non predictable way!
Good find!

PR #568, opened 2020-02-28
Ability to alter a column and set NOT NULL

eolivelli:
- Allow to alter a column and set NOT NULL
- Allow to set NULL to a NOT NULL column
- Check that the column does not contain NULLs in case of change to "NOT NULL"

The check is not blocking the table, we have to improve it in the future.

### DIFF
--- a/herddb-core/src/main/java/herddb/core/TableManager.java
+++ b/herddb-core/src/main/java/herddb/core/TableManager.java
@@ -39,6 +39,7 @@ import herddb.model.ColumnTypes;
 import herddb.model.DDLException;
 import herddb.model.DMLStatementExecutionResult;
 import herddb.model.DataScanner;
+import herddb.model.DataScannerException;
 import herddb.model.DuplicatePrimaryKeyException;
 import herddb.model.GetResult;
 import herddb.model.Index;
@@ -48,6 +49,7 @@ import herddb.model.Record;
 import herddb.model.RecordFunction;
 import herddb.model.RecordTooBigException;
 import herddb.model.ScanLimits;
+import herddb.model.ScanLimitsImpl;
 import herddb.model.Statement;
 import herddb.model.StatementEvaluationContext;
 import herddb.model.StatementExecutionException;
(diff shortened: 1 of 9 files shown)


eolivelli:
@amitvc you worked on this stuff recently.
If you have time please take a look.
We need this patch for next release of EmailSuccess

herddb-core/src/main/java/herddb/sql/DDLSQLPlanner.java
```diff
@@ -886,17 +890,24 @@ private Statement buildAlterStatement(String defaultTableSpace, Alter alter) thr
...
+                            columnSpecs
                     );
 
                     if (oldColumn.type != newType) {
-                        throw new StatementExecutionException("cannot change datatype to " + cl.getColDataType().
-                                getDataType()
-                                + " for column " + columnName + " in table " + tableName + " in tablespace '" + tableSpace + "'");
+                        if (ColumnTypes.isNotNullToNullConversion(oldColumn.type, newType)) {
+                            // allow change from "STRING NOT NULL" to "STRING NULL"
+                        } else if (ColumnTypes.isNullToNotNullConversion(oldColumn.type, newType)) {
```
amitvc: For the null to NOT null use case what happens if the table already has null values for the columns?

herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -3529,6 +3531,46 @@ public boolean isStarted() {
...
+            // fast fail
+            scan.setLimits(new ScanLimitsImpl(1, 0));
+            boolean foundOneNull = false;
+            try (DataScanner scanner = this.scan(scan, context, null, false, false);) {
+                foundOneNull = scanner.hasNext();
+            } catch (DataScannerException err) {
+                throw new StatementExecutionException(err);
+            }
+            if (foundOneNull) {
+                throw new StatementExecutionException("Found a record in table " + table.name + " that contains a NULL value for column " + column + " ALTER command is not possible");
```
amitvc: I see this is where you are doing the null check on the column values. For a large table will this operation be a problem? This happens async?

eolivelli: This is blocking.
The alter statement will complete only after this code has been fully executed

We aren't blocking the table explicitly so actually we could have problems if another client is modifying the table concurrently.

Blocking the table, like acquiring a global write (checkpoint) lock is doable but I did not do it because it should be implemented carefully.

The immediate need is to alter table schema while the application is stopped so this is the minimal change.

We should fix the concurrent access case in a follow up case.

Currently (before this patch) the implementation is already broken because if you try to change a column from not null to null able the alter command succeeds but it really does not have effect (the column stays nullable!)

eolivelli:
The follower replay the full story of the table so the answer to your question is 'yes'


I feel there is no need for an additional test

Please help merging  if you are okay

Thank you for your review

eolivelli:
We should rekick travis

amitvc:
@eolivelli Should I merge it all checks have passed successfully.

eolivelli:
Yes @amitvc please help merging
All tests are green now

PR #572, opened 2020-03-04
Write ledger header

eolivelli:
- write an entry to the ledger before adding it to the list of active ledgers
- this is a workaround to a possible fault case in BookKeeper when there is no entry on the Bookie and one Bookie goes down


eolivelli:
The attached test case fails with this fix and reproduces the problems seen by @hamadodene in his test environment

PR #573, opened 2020-03-05
Retry connection and refresh metadata cache on timeout failures during connection opening

diegosalvi:
On timeout failures during performAuthentication the attempt is totaly aborted but cached routes aren refreshed. In such cases remote server opened the connection but still cannot answer the client about leadership state. If the server lost leadership on a tablespace the client will still attempt forever to connect there instead checking the right tablespace leader.

In addition used already existing exception HDBOperationTimeoutException to mark timeout exceptions during normal executions and a new exception ConnectionOpenTimeoutException to mark timeout failure while performing connection opening


eolivelli:
I feel I have addressed this problem with
https://github.com/diennea/herddb/commit/1595f7ed6a9d4e8586ef64a437dbfc250aad9451

What's the error you saw that you are fixing with this patch ?

diegosalvi:
This PR in some way gone in the wrong direction... I will submit a simpler pr

PR #574, opened 2020-03-05
Retry connection and refresh metadata cache on timeout failures during connection opening

diegosalvi:
On timeout failures during performAuthentication the attempt is totaly aborted but cached routes aren refreshed. In such cases remote server opened the connection but still cannot answer the client about leadership state. If the server lost leadership on a tablespace the client will still attempt forever to connect there instead checking the right tablespace leader.

Added a testcase (it fails without this pr)

### DIFF
--- a/herddb-core/src/main/java/herddb/client/HDBConnection.java
+++ b/herddb-core/src/main/java/herddb/client/HDBConnection.java
@@ -415,7 +415,7 @@ public class HDBConnection implements AutoCloseable {
         }
     }
 
-    public RoutedClientSideConnection getRouteToTableSpace(String tableSpace) throws ClientSideMetadataProviderException, HDBException {
+    protected RoutedClientSideConnection getRouteToTableSpace(String tableSpace) throws ClientSideMetadataProviderException, HDBException {
         if (closed) {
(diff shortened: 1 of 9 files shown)


herddb-core/src/main/java/herddb/client/HDBConnection.java
```diff
@@ -415,7 +415,7 @@ private RoutedClientSideConnection getRouteToServer(String nodeId) throws Client
         }
     }
 
-    private RoutedClientSideConnection getRouteToTableSpace(String tableSpace) throws ClientSideMetadataProviderException, HDBException {
+    public RoutedClientSideConnection getRouteToTableSpace(String tableSpace) throws ClientSideMetadataProviderException, HDBException {
```
eolivelli: this should not  be exposed to the public API, or at least we should add a javadoc that tells not to use it

diegosalvi: Reduced to protected and handled public visibility in test

eolivelli:
This test is stuck on Travis
[INFO] Running herddb.server.SimpleClientServerTest

PR #580, opened 2020-03-15
Implement UPSERT

### LINKED ISSUE #579: Add "PUT" operation - UPSERTS
Is this a question, a feature request or a bug report?

**FEATURE REQUEST**

1. Please describe the feature you are requesting.
In order to ease bulk loads it would be great to have the ability to perform an INSERT operation that does not fail.
In case you are trying to insert a record that is totally equivalent (same logical content) to an existing one no action is written to the WAL.
In case you are updating the value, an UPDATE operation will be written to the WAL.

  
2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

should-have

eolivelli:
- implement basic experimental UPSERT support
- use UPSERT in HerdDB Collections framework
- the UPSERT follows exactly the same code path of INSERT, but in case of DuplicatePrimaryKey it turns the INSERT to a full UPDATE, of the full record

Fixes #579


eolivelli:
Unfortunately Calcite treats an UPSERT like an INSERT.
We are temporary trying to detect only the 'UPSERT' all uppercase Key word.
We should add support in Calcite.
By now this fix is enough for us in order to leverage this powerful feature

herddb-core/src/test/java/herddb/core/UpdateTest.java
```diff
@@ -183,4 +186,43 @@ public void updateMultiRowsWithValidationError() throws Exception {
...
+                assertEquals(1235, recordSet.get(0).get(0));
+            }
+            assertEquals(4, executeUpdate(manager, "UPSERT INTO tblspace1.tsql(k1,n1)"
+                    + "values(?,?),(?,?),(?,?),(?,?)", Arrays.asList(
+                            "mykey", Integer.valueOf(1235),
+                            "mykey", Integer.valueOf(1236),
+                            "mykey", Integer.valueOf(1237),
+                            "mykey", Integer.valueOf(1238)
+                            )).getUpdateCount());
+            try (DataScanner scan = scan(manager, "SELECT n1 from tblspace1.tsql where k1=?", Arrays.asList("mykey"))) {
```
amitvc: If you call upsert on a table with non-null fields will it throw an exception regarding non null values? And if null values are allowed it will insert whatever is provided?

eolivelli: It works exactly like an INSERT, falling back to a FULL record UPDATE.
So handling of null values will behave as usual.

Do you think it is worth to add an explicit test case?

eolivelli: @amitvc  I have added more tests cases.

eolivelli:
@amitvc I have added more test cases.

Please help merging this patch if it looks good after my last commit.

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
```diff
@@ -347,6 +349,12 @@ public TranslatedQuery translate(String defaultTableSpace, String query, List<Ob
         }
     }
 
+    private static boolean detectUpsert(String query) {
+        // unfortunately Calcite does not retain the UPSERT flag in LogicalTableModify
+        boolean upsert = query.startsWith("UPSERT");
```
diegosalvi: We should open another issue to fix this

PR #581, opened 2020-03-18
Update ZooKeeper to 3.6.1

eolivelli:
Switch to ZK 3.6.1 client

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 .idea
 **/*.iml
 *.iml
+**/.DS_Store
 
 # Release
 pom.xml.tag
--- a/herddb-core/pom.xml
+++ b/herddb-core/pom.xml
@@ -55,6 +55,18 @@
             <groupId>org.apache.zookeeper</groupId>
             <artifactId>zookeeper-jute</artifactId>
         </dependency>
+        <dependency>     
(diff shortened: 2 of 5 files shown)


eolivelli:
@mino181295  we should make a similar fix in BookKeeper Visual Manager.
Because ZooKeeper Server needs Snappy and Dropwizard on the class path.

eolivelli:
see https://github.com/apache/bookkeeper/issues/2292

mino181295:
I think curator does not support Zookeeper 3.6.0 in 4.x

It will be supported in 5
https://github.com/apache/curator/commit/7bbb7196173b88baac0e102bdff265dc7f2ee77e

eolivelli:
We are not using Curator @mino181295 in this project.
If we are using it in BKVM we can drop it, maybe it is only for tests

eolivelli:
It looks like that with ZK 6.3.1 we do not have any more the problem with BK
https://github.com/apache/zookeeper/pull/1287

cc @dmercuriali @mino181295

eolivelli:
@mino181295 @amitvc @diegosalvi this patch is ready for merge

mino181295:
LGTM

eolivelli:
thank you @mino181295 
I am merging this change now

PR #588, opened 2020-04-09
HerdDB Collections Improvements: clear and better ValueSerializer

eolivelli:
- supports clear() (TRUNCATE TABLE)
- let the ValueSerializer return directly a byte[] in order to save memory and CPU cycles

### DIFF
--- a/herddb-collections/src/main/java/herddb/collections/CollectionsManager.java
+++ b/herddb-collections/src/main/java/herddb/collections/CollectionsManager.java
@@ -73,7 +73,7 @@ public final class CollectionsManager implements AutoCloseable {
                 try (VisibleByteArrayOutputStream oo = new VisibleByteArrayOutputStream(expectedSize);
                         ObjectOutputStream ooo = new ObjectOutputStream(oo)) {
                     ooo.writeUnshared(object);
-                    return oo.toByteArray();
+                    return oo.toByteArrayNoCopy();
                 }
             }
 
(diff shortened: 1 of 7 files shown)


herddb-collections/src/main/java/herddb/collections/CollectionsManager.java
```diff
@@ -66,24 +66,28 @@
...
-            try (ObjectOutputStream ooo = new ObjectOutputStream(oo)) {
-                ooo.writeUnshared(object);
+    private static ValueSerializer defaultValueSerializer(int expectedSize) {
+        return new ValueSerializer() {
+            @Override
+            public byte[] serialize(Object object) throws Exception {
+                try (VisibleByteArrayOutputStream oo = new VisibleByteArrayOutputStream(expectedSize);
+                        ObjectOutputStream ooo = new ObjectOutputStream(oo)) {
+                    ooo.writeUnshared(object);
+                    return oo.toByteArray();
```
diegosalvi, line 76: Easy improvement. ToByteArray only if expected and real size differs otherwise getBuffer whould be better due to zero copy. We could even add a method with such logic

diegosalvi:
+1, just a note

eolivelli:
@diegosalvi  I have addressed your comment

diegosalvi:
Good.. I like copy less byte arrays :D

Waiting CI

Issue #592, opened 2020-04-27
HerdDB Collection: ability to not create row-level locks


eolivelli:
For HerdDB Collections users most of the time row-level locks are only a waste of CPU/memory as most of the time one single table is accessed by the only single thread that created it.

We should provide a way to disable row-level locks, like a "enableConcurrentAccess(false)" on TmpMap builder.
By default TmpMap should be thread safe, otherwise we will break existing applications.

PR #594, opened 2020-04-28
HerdDB Collection: ability to not create row-level locks #592

pv3nturi:
Added the new ServerConfiguration property _**server.tablemanager.disablerowlevellocks**_ (default **false**) to not/perform db writes/reads in thread-safe mode.

Added the new method  _**threadsafe(boolean)**_ for the CollectionsManager builder in order to enable/disable concurrent-access for all TmpMaps.

Removed useless LegacyLocalLockManager.


herddb-collections/src/main/java/herddb/collections/CollectionsManager.java
```diff
@@ -348,6 +353,16 @@ public CollectionsManager build() {
             return this;
         }
 
+        /**
+         * Whether all tables access have performed by single thread concurrent access management may be
+         * useless and lead to bad performance. By noLock all operations will be performed in "thread-unsafe mode".
+         * @return the builder itself
+         */
+        public TmpMapBuilder<V1> withNoLock() {
```
eolivelli, line 361: allowConcurrentAccess(boolean enabled)

herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -184,8 +181,7 @@
     /**
      * Local locks
      */
-    private final ILocalLockManager locksManager =
-            USE_LEGACY_LOCK_MANAGER ? new LegacyLocalLockManager() : new LocalLockManager();
+    private ILocalLockManager locksManager = new LocalLockManager();
```
eolivelli, line 184: please leave it final, it is better for the JVM for optimizations and in order to prevent NPEs

herddb-utils/src/main/java/herddb/utils/LocalNoLockManager.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+package herddb.utils;
+
+/**
+ * Handle fake locks by key in order to avoid useless concurrent access management.
+ *
+ * @author paolo.venturi
+ */
+public class LocalNoLockManager implements ILocalLockManager {
```
eolivelli, line 28: NullLockManager?

herddb-utils/src/main/java/herddb/utils/LocalNoLockManager.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public LockHandle acquireWriteLockForKey(Bytes key) {
+        return new LockHandle(0, key, true);
```
eolivelli, line 32: please do not create a new instance every time, we will add pressure on GC.

eolivelli: is it an option to return "null" ?

herddb-core/src/main/java/herddb/model/Transaction.java
```diff
@@ -128,6 +128,9 @@ public LockHandle lookupLock(String tableName, Bytes key) {
     }
 
     public void registerLockOnTable(String tableName, LockHandle handle) {
+        if (handle == null) {
```
eolivelli, line 131: this is not good

pv3nturi: Needed to let LockHandle in NullLockManager to be null

eolivelli: This can lead to errors in core execution paths.
It is better to have a dummy LockHandle, only one singleton instance

herddb-core/src/main/java/herddb/server/ServerConfiguration.java
```diff
@@ -213,6 +213,13 @@
     public static final int PROPERTY_PORT_AUTODISCOVERY = 0;
     public static final int PROPERTY_ZOOKEEPER_SESSIONTIMEOUT_DEFAULT = 40000;
 
+    /**
+     * For HerdDB Collections users most of the time row-level locks are only a waste of CPU/memory
+     * as most of the time one single table is accessed by the only single thread that created it.
+     */
+    public static final String PROPERTY_TABLEMANAGER_CONCURRENT_ACCESS = "server.tablemanager.concurrentaccess";
```
eolivelli, line 220: It is better to name this property:
PROPERTY_TABLEMANAGER_DISABLE_ROWLEVELLOCKS = "server.tablemanager.disablerowlevellocks"

@diegosalvi  @amitvc  @hamadodene  it would be useful in the future to work this way on the "Follower" because on Follower nodes there is no concurrent access to the same record

herddb-core/src/main/java/herddb/model/Transaction.java
```diff
@@ -128,6 +128,9 @@ public LockHandle lookupLock(String tableName, Bytes key) {
     }
 
     public void registerLockOnTable(String tableName, LockHandle handle) {
+        if (handle.key == null) {
```
eolivelli, line 131: this is not needed, please drop, useless code

PR #597, opened 2020-05-02
Diskless cluster mode - first implementation

eolivelli:
This is the implementation of a new DataStorageManager totally based on BookKeeper + ZooKeeper.: BookKeeperDataStorageManager

It is supposed to be used in pure cloud environment when you already have a ZooKeeper + BookKeeper cluster and you want to store all of the data on those pods.

A valid use case is for Apache Pulsar Manager, when you already have a BK cluster for Pulsar.

It is expected not to start the Bookie in  "embedded bookie" when working on diskless cluster mode, because it won't make sense at all, using local disk is better if you have a stateful pod for HerdDB it is better to use the default FileDataStorageManager

In order to activate diskless cluster mode you only have to configure:
- server.mode=diskless-cluster
- server.node.id=MYID
- configure ZK settings

server.node.id is required because disk is expected to be ephemeral and we must have a persistent node id.
Data pages and Index pages are stored on BookKeeper, one ledger per page.
Metadata about mapping between HerdDB pages and BK ledgers are stored on ZK.
Every other kind of data and metadata is persisted on ZooKeeper:
- table and index status
- checkpoint information
- temporary transaction during checkpoints

ZK znodes have an hard limit in size, so in the future we have to improve the storage for potentially big data stored on znodes, like the transactions persisted during checkpoint.

Garbage collection works are for all of the other storage engines.

It is expected to use this mode when:
- you do not have persisted disks for HerdDB but you have a BookKeeper and ZooKeeper cluster
- database mostly fits into memory, or you are mostly performing writes over reads
- you have only one HerdDB server (you can still have a replication factor bigger than 1)

Replication:
- replication works as usual, but when you are writing data pages/index pages, data is replicated on bookies
- the replication factor for data is the same as for the WAL
- each server has its own data space on BK + ZK, if you start more servers each server will have its own copy of data, replicated on BK according to the replication factor

It is not expected to run two instances of the same server (identified by the node id), exactly as you do with other modes. Each instance of the server may overwrite data pages leading to data corruption.
This is something we should improve in the future, but it is not a trivial task.

### DIFF
--- a/herddb-core/src/main/java/herddb/cluster/BookKeeperDataStorageManager.java
+++ b/herddb-core/src/main/java/herddb/cluster/BookKeeperDataStorageManager.java
@@ -54,12 +54,14 @@ import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -72,10 +74,12 @@ import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import org.apache.bookkeeper.client.BKException;
 import org.apache.bookkeeper.client.BKException.BKNoSuchLedgerExistsOnMetadataServerException;
+import org.apache.bookkeeper.client.api.DigestType;
 import org.apache.bookkeeper.client.api.LedgerEntries;
 import org.apache.bookkeeper.client.api.ReadHandle;
 import org.apache.bookkeeper.client.api.WriteHandle;
 import org.apache.bookkeeper.common.concurrent.FutureUtils;
+import org.apache.bookkeeper.stats.Counter;
 import org.apache.bookkeeper.stats.NullStatsLogger;
 import org.apache.bookkeeper.stats.OpStatsLogger;
 import org.apache.bookkeeper.stats.StatsLogger;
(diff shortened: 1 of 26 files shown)


amitvc:
@eolivelli Can you please add some context to this ticket? So it's helpful to study the code changes.

eolivelli:
It is an early prototype.
Basically I have copied FileDataStorageManager replacing all metadata persistence functions with ZK and all data pages are written to BK.

It is mostly a refactor and copy and paste from the other class.
I have to add tests that cover all functions, possibly leveraging existing FileDataStorageManager tests.

I will also have to plug it into Server bootstrap code

The idea is that all that you were writing to local disk now it will be stored on ZK + BK

See the linked issue on Pulsar Manager

eolivelli:
@amitvc I have added a detailed description.
@diegosalvi  PTAL

herddb-core/src/main/java/herddb/cluster/BookKeeperDataStorageManager.java
```diff
@@ -0,0 +1,1744 @@
...
+                }
+            }
+
+        } catch (IOException | org.apache.bookkeeper.client.api.BKException err) {
+            throw new DataStorageManagerException(err);
+        } catch (InterruptedException err) {
+            Thread.currentThread().interrupt();
+            throw new DataStorageManagerException(err);
+        }
+        getTableSpacePagesMapping(tableSpace).getIndexPagesMapping(indexName).writePageId(pageId, ledgerId);
```
diegosalvi, line 1105: BLink trees keep a two page id sets: one is the flush set and can be overridden the other is the checkpoint set and is never overridden. Both use writeIndexPage then here we could overwrite page to ledger knowledge and forget the old ledger (and miss his deletion on a future checkpoint when the ledger itself won't be needed anymore)

eolivelli:
Unfortunately there is a problem on CI.
I must check

eolivelli:
CI is green
@diegosalvi  PTAL and help me merging this patch

diegosalvi:
Great work

PR #626, opened 2020-06-22
Records updates are not visible to current transaction in scan for records born within the transaction

eolivelli:


herddb-core/src/main/java/herddb/model/Transaction.java
```diff
@@ -235,19 +235,31 @@ public void unregisterUpgradedLocksOnTable(String tableName, LockHandle lock) {
...
-        if (ll == null) {
-            ll = new HashMap<>();
-            changedRecords.put(tableName, ll);
+            // this return value is not important
+            return false;
         }
-        if (value == null) {
-            ll.remove(key);
+        Map<Bytes, Record> newRecordsForTable = newRecords.get(tableName);
+        if (newRecordsForTable != null && newRecordsForTable.containsKey(key)) {
```
diegosalvi, line 244: Just a consideration: Could it be better with a Map.compute to avoid the double key lookup?

eolivelli: > Just a consideration: Could it be better with a Map.compute to avoid the double key lookup?

I am not sure I understand what you mean here.
We have to enter this branch only if the key is present in `newRecordsForTable`
otherwise we have to enter the other branch.

diegosalvi: Let this be as is

PR #637, opened 2020-06-26
Add OpenJPA Compatibility

eolivelli:
- use a global register for datasources created by the JDBC Driver
- canonicalize the key for each datasource, as OpenJPA calls 'connect' with different Properties objects
- add minimal test case
- implement a few HerdDBResultSet methods
- implement jdbc driver information
- deregistering the driver will close all of the datasources, datatasources that started the embedded server will stop the server
- implement "autoClose" JDBC property, that requests the JDBC driver to autoClose the datasource and stop the embedded database when there is any active connection anymore
- add new OpenJPA DBDictionary for HerdDB in order to tell OpenJPA to escape every identifier and to let it know (un)supported capabilities

Fixes #634

PR #638, opened 2020-06-27
Basic implementation for 'FOR UPDATE' clause

### LINKED ISSUE #631: FOR UPDATE not supported
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

Standard SQL support FOR UPDATE, ex:

    SELECT SEQUENCE_VALUE FROM OPENJPA_SEQUENCE_TABLE WHERE ID = ? FOR UPDATE

HerdDB does not.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

This prevents to use several JPA mapper without workarounds.

eolivelli:
Implementation of SELECT .... FOR UPDATE.

If you issue "SELECT .... FOR UPDATE" the transaction will hold "WRITE" locks for every record returned by the query (both SCANs and GETs).

Traditionally if you do not add "FOR UPDATE" the transaction holds only "READ" locks for every record return by the query, this behaviour is not changed.

Unfortunately the detection of "FOR UPDATE" syntax is quite hacky, because Apache Calcite does not still implement it.

This implementation will ease the compatibility with JPA providers

fixes #631 

Once #637 is in we can change the OpenJPA test case and let it use "FOR UPDATE" syntax

### DIFF
--- a/herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
+++ b/herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
@@ -243,10 +243,18 @@ public class CalcitePlanner implements AbstractSQLPlanner {
                 + ",query:" + query
                 + ",returnValues:" + returnValues
                 + ",maxRows:" + maxRows;
+        boolean forceAcquireWriteLock;
+        if (query.endsWith(" FOR UPDATE") // this looks very hacky
+                && query.substring(0, 6).toLowerCase().equals("select")) {
+            forceAcquireWriteLock = true;
+            query = query.substring(0, query.length() - " FOR UPDATE".length());
+        } else {
+            forceAcquireWriteLock = false;
+        }
         if (allowCache) {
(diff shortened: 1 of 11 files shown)


eolivelli:
@mino181295 this change should also fix your problem with EclipseLink using auto generated keys

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
```diff
@@ -243,10 +243,18 @@ public TranslatedQuery translate(String defaultTableSpace, String query, List<Ob
                 + ",query:" + query
                 + ",returnValues:" + returnValues
                 + ",maxRows:" + maxRows;
+        boolean forceAcquireWriteLock;
+        if (query.endsWith(" FOR UPDATE") // this looks very hacky
```
diegosalvi, line 247: Really hacky unfortunately :(

PR #642, opened 2020-07-02
Update jSQLParser to 3.2

eolivelli:
Update jSQLParser to 3.2
https://github.com/JSQLParser/JSqlParser/releases

suggested by @rmannibucau in order to enhance compatibility with OpenJPA


eolivelli:
@diegosalvi  @mino181295 this change may break Apache Pulsar Manager, that uses MyIBatis (that uses jSQLParser) and https://magnews.com, that uses jSQLParser internally.

PR #643, opened 2020-07-02
Do not pool connections in JDBC Driver by default

eolivelli:
- introduce property poolConnections to specify use of Commons Pool for JDBC DataSource
- enable it by default for DataSources
- disable it by default for JDBC Driver

The idea is that users of the pure JDBC Driver normally use their own pool, like https://github.com/brettwooldridge/HikariCP, there is no need to add the cost of a useless pool

### DIFF
--- a/herddb-jdbc/src/main/java/herddb/jdbc/BasicHerdDBDataSource.java
+++ b/herddb-jdbc/src/main/java/herddb/jdbc/BasicHerdDBDataSource.java
@@ -63,6 +63,7 @@ public class BasicHerdDBDataSource implements javax.sql.DataSource, AutoCloseabl
     private int waitForTableSpaceTimeout = 60000;
     private boolean discoverTableSpaceFromQuery = true;
     private HDBConnection connection;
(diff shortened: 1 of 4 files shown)


herddb-jdbc/src/main/java/herddb/jdbc/BasicHerdDBDataSource.java
```diff
@@ -163,6 +164,14 @@ public synchronized void setUrl(String url) {
         this.url = url;
     }
 
+    public synchronized boolean isPoolConnections() {
+        return poolConnections;
+    }
+
+    public synchronized void setPoolConnections(boolean poolConnections) {
```
diegosalvi, line 171: IMHO to be cleaner this shouldn't be settable is client != null to not mix pooled/non pooled connection behavior (sure is a super edge and buggy case)

eolivelli: done

PR #644, opened 2020-07-02
Allow multiple 'local' databases: jdbc:herddb:local:db1 syntax

### LINKED ISSUE #639: support to name local database
Feature request:

:local mode is neat in tests but when having multiple datasources it would be great to be able to append a name:

    jdbc:herddb:local:myds

Enrico explained be it is already supported with a "remote" syntax with parameters but I think it makes sense to just make it built-in.

eolivelli:
Fixes #639

### DIFF
--- a/herddb-net/src/main/java/herddb/network/netty/NettyChannelAcceptor.java
+++ b/herddb-net/src/main/java/herddb/network/netty/NettyChannelAcceptor.java
@@ -249,7 +249,7 @@ public class NettyChannelAcceptor implements AutoCloseable {
         });
         InetSocketAddress address = new InetSocketAddress(host, port);
         LOGGER.log(Level.INFO, "Starting HerdDB network server at {0}:{1}", new Object[]{host, port + ""});
-        if (address.isUnresolved()) {
+        if (enableRealNetwork && address.isUnresolved()) {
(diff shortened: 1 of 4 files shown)


eolivelli:
I will update the wiki as soon as this PR is merged

herddb-net/src/main/java/herddb/network/netty/NettyChannelAcceptor.java
```diff
@@ -249,7 +249,7 @@ public Integer getSample() {
         });
         InetSocketAddress address = new InetSocketAddress(host, port);
         LOGGER.log(Level.INFO, "Starting HerdDB network server at {0}:{1}", new Object[]{host, port + ""});
-        if (address.isUnresolved()) {
+        if (enableRealNetwork && address.isUnresolved()) {
```
diegosalvi, line 252: Is this related to multiple local databases?

eolivelli: yes, because we are going to use the name of the database as "hostname".
jdbc:herddb:local:mydb

as we are using Netty Local Channel we won't need this to be a valid hostname.
we need it to be a resolved hostname when we are using real network and especially when you are using Kerberos based auth

so in case of "local" we can skip that check and we must do it otherwise it will fail

PR #645, opened 2020-07-02
Implement DEFAULT VALUES

eolivelli:
- Implement "DEFAULT" values for basic data types, all apart from byte[].
- for Timestamp columns we are supporting only CURRENT_STATEMENT
- add default_value to syscolumns system table
- add COLUMN_DEF metadata to JDBC Driver (compatible with older servers)
- implement serialization of Table metadata (we can read metadata of older versions, if you create a table with DEFAULT values, you won't be able to rollback the serverto previous version)
- the default value is applied by Calcite, that injects default values into the INSERT statement
- UPSERT works like an INSERT, so it applies all default values
- support default value in "SHOW CREATE TABLE" command
- implement DEFAULT in "ALTER TABLE MODIFY COLUMN" (modify(
- implement DEFAULT in "ALTER TABLE CHANGE COLUMN"  (rename and/or change type)
- implement DEFAULT in "ALTER TABLE ADD" (new column)

Risks:
previously we ignored any kind of 'DEFAULT' spec on columns, applications that ported SQL DDL code from MySQL to HerdDB will see a behaviour change


eolivelli:
This patch is only a draft, I have to add more test cases !

I have also found a bug on "boolean" types representation, I guess no one ever used a boolean column ! because it did not work at all (bug on Bytes serialization, that swapped true with false!)

herddb-core/src/main/java/herddb/model/Table.java
```diff
@@ -176,13 +179,17 @@ public static Table deserialize(byte[] data) {
             for (int i = 0; i < ncols; i++) {
                 long cversion = dii.readVLong(); // version
                 long cflags = dii.readVLong(); // flags for future implementations
-                if (cversion != 1 || cflags != 0) {
+                if (cversion != 1 || (cflags != 0 && cflags != COLUMNFLAGS_HAS_DEFAULT_VALUE)) {
```
diegosalvi, line 182: Couldn't just be (cversion != 1 || cflags != COLUMNFLAGS_HAS_DEFAULT_VALUE)?

herddb-core/src/main/java/herddb/model/Table.java
```diff
@@ -176,13 +179,17 @@ public static Table deserialize(byte[] data) {
...
                     throw new IOException("corrupted table file");
                 }
                 String cname = dii.readUTF();
                 int type = dii.readVInt();
                 int serialPosition = dii.readVInt();
-                columns[i] = Column.column(cname, type, serialPosition);
+                Bytes defaultValue = null;
+                if (cflags == 1) {
```
diegosalvi, line 189: Use constant COLUMNFLAGS_HAS_DEFAULT_VALUE

herddb-core/src/main/java/herddb/model/Table.java
eolivelli: actually allowed values are:
0 -> NO FLAG
1 -> COLUMNFLAGS_HAS_DEFAULT_VALUE

we still have to allow cflags == 0 in order to be compatible  with old versions.
we are now still keeping flags = 0 if there is no default value, in order to allow rollbacks to previous versions in case you did not use the new DEFAULT VALUE feature

btw I have introduced constants

herddb-core/src/main/java/herddb/model/Table.java
eolivelli: done.
I am using now bitwise comparison, this way it is clearer the meaning of the flag for future extensions

eolivelli:
@diegosalvi I have addressed your comments

eolivelli:
@diegosalvi I have resolved the conflicts

Issue #625, opened 2020-06-22
Implement a way to create secondary indexes on nullable values


eolivelli:
Currently, in 0.16.x, you cannot index a column that allow null values.

We should support this kind of index keys, but we must preserve semantics.

In order to make it simple we could simply just NOT index records that contain NULL values for indexed columns, but we won't be able to implement correctly the semantics for multi column indexes (that is the case of the customer)

PR #646, opened 2020-07-03
Issue 625: Allow indexing of NULL values for single column indexes

eolivelli:
- allow NULLs on columns indexed by single column indexes
- multi column indexes still do not support null values
- "IS NULL" and "IS NOT NULL" clauses does not take indexes into account, the index contains only non-null values

Relates to #625


eolivelli:
This is only a draft, it deserves more tests

@rmannibucau you can try your tests using this branch.
I will finish the work during the weekend

rmannibucau:
@eolivelli once the backquote issue fixed it runs slowly but perfectly on my test app

PR #648, opened 2020-07-05
Add support for not null for timestamp, double, boolean, byte fields

amitvc:
Tests - 
I have added unit tests to ensure that not null for the timestamp, double, boolean, byte fields work.
Also ran all unit tests on the project to ensure nothing is broken.

### DIFF
--- a/herddb-core/src/test/java/herddb/core/SystemTablesTest.java
+++ b/herddb-core/src/test/java/herddb/core/SystemTablesTest.java
@@ -426,17 +426,17 @@ public class SystemTablesTest {
         }
     }
 
-    @Test(expected = StatementExecutionException.class)
-    public void testForNonSupportedTypesWithNotNullConstraints() throws Exception {
-        String nodeId = "localhost";
-        try (DBManager manager = new DBManager(nodeId, new MemoryMetadataStorageManager(), new MemoryDataStorageManager(), new MemoryCommitLogManager(), null, null)) {
-            manager.start();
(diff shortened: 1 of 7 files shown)


herddb-core/src/test/java/herddb/core/SystemTablesTest.java
```diff
@@ -426,17 +426,17 @@ public void testSystemTablesForNonNullColumnTypes() throws Exception {
         }
     }
 
-    @Test(expected = StatementExecutionException.class)
```
eolivelli, line 429: Can you please drop these lines instead of commenting them?

herddb-utils/src/main/java/herddb/utils/Bytes.java
```diff
@@ -347,7 +347,7 @@ public static int toInt(byte[] array, int index) {
     }
 
     public static boolean toBoolean(byte[] bytes, int offset) {
-        return bytes[offset] != 1;
+        return bytes[offset] == 1;
```
eolivelli, line 350: Good catch!

eolivelli:
Please also very checkstyle, check on Travis or run
mvn checkstyle:check

herddb-core/src/test/java/herddb/core/SystemTablesTest.java
amitvc: Yes. Will do.

herddb-utils/src/main/java/herddb/utils/Bytes.java
amitvc: Thanks 👍

eolivelli:
committed to master branch, thank you @amitvc  !

PR #656, opened 2020-07-09
SELECTS that contain the 'from' string does not work with tablespace autodiscovery

eolivelli:
TableSpace autodetection does not work in these cases:
- UPSERT statements
- statements like `SELECT COUNT(M.id) FROM myts.a M WHERE 1=1 AND (M.id IN (SELECT H.id FROM myts.b AS H WHERE H.FIELD_NAME='From' AND H.FIELD_VALUE LIKE 'aaa'))`

The problem is that the query contains a string literal 'from'

### DIFF
--- a/herddb-core/src/test/java/herddb/utils/QueryUtilsTest.java
+++ b/herddb-core/src/test/java/herddb/utils/QueryUtilsTest.java
@@ -64,7 +64,7 @@ public class QueryUtilsTest {
         assertEquals(theTableSpace, QueryUtils.discoverTablespace(defaultTableSpace, "insert into myts.test values(?,?,?,?)"));
         assertEquals(theTableSpace, QueryUtils.discoverTablespace(defaultTableSpace, "INSERT INTO myts.a SELECT * FROM myts.c M WHERE 1=1 AND (M.id IN (SELECT H.id FROM myts.b AS H WHERE H.FIELD_NAME='Into' AND H.FIELD_VALUE LIKE 'aaa'))"));
     }
- 
+
     @Test
     public void testDiscoverTablespaceFromUpsert() {
         String defaultTableSpace = TableSpace.DEFAULT;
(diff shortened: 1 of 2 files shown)


eolivelli:
I haven't found a good fix for the 'SELECT' case.
Anyone has an idea about how to fix the REGEX ?

aluccaroni:
@eolivelli something like `(?i)select.+[^a-zA-Z_0-9'"]+from[^a-zA-Z_0-9'"]+` ?

eolivelli:
@aluccaroni 
I am sorry but I don't know how to apply your suggestion here:
https://github.com/diennea/herddb/blob/7fa395c8ececf3ea25fc90a75fd8e6bb9cdf64f7/herddb-core/src/main/java/herddb/utils/QueryUtils.java#L38
Do you want to checkout this branch and pickup this PR ?

If the QueryUtilsTest passes than you have found the right medicine

Personally I hate regexps :-)

herddb-core/src/test/java/herddb/utils/QueryUtilsTest.java
```diff
@@ -66,6 +66,7 @@ public void testDiscoverTablespaceFromInsert() {
         assertEquals(defaultTableSpace, QueryUtils.discoverTablespace(defaultTableSpace, "insert into test values(?,?,?)"));
         assertEquals(theTableSpace, QueryUtils.discoverTablespace(defaultTableSpace, "insert into myts.test(a,b) values(?,?,?,?)"));
         assertEquals(theTableSpace, QueryUtils.discoverTablespace(defaultTableSpace, "insert into myts.test values(?,?,?,?)"));
+	assertEquals(theTableSpace, QueryUtils.discoverTablespace(defaultTableSpace, "INSERT INTO myts.a M WHERE 1=1 AND (M.id IN (SELECT H.id FROM myts.b AS H WHERE H.FIELD_NAME='Into' AND H.FIELD_VALUE LIKE 'aaa'))"));
```
eolivelli, line 69: probably this SQL does not make sense
"INSERT INTO myts.a M WHERE"

what about
`"INSERT INTO myts.a SELECT * FROM myts.c M WHERE 1=1 AND (M.id IN (SELECT H.id FROM myts.b AS H WHERE H.FIELD_NAME='Into' AND H.FIELD_VALUE LIKE 'aaa'))"`

aluccaroni: yeah, I went in blind ;-)

PR #657, opened 2020-07-10
Allow NULL in multi-column secondary indexes

### LINKED ISSUE #625: Implement a way to create secondary indexes on nullable values

eolivelli:
- Allow NULL values in multi-column secondary indexes
- If the first column is NULL the record is not indexed
- Serialize other columns up to the first NULL, in order to allow SecondaryIndexPrefixScans
- Add tests about RecordSerializer
- Clean up default buffer sizes (4kb is too much for keys)
- reorganize `compare(Object a, Object b)` in order to reduce the number of "instanceof" (we can't get rid of them)

Closes #625


diegosalvi:
+1 for reducing buffer size

PR #658, opened 2020-07-15
Implement more JDBC features

eolivelli:
Implement more JDBC features:
- java.sql.Time and java.sql.Date (as dummy wrapper for java.sql.Timestamp!)
- add aliases for "time with timezone","time" and "date", mapped to "timestamp" (we are not dealing with datetime stuff without timezone yet)
- make JDBC metadata non case sensitive and force returned table_names and names in general to be lowercase, in order to guarantee current behavior in case of future changes to the server side behavior
- implement getCursorName (mostly useless, but with no cost at runtime)
- throw SQLFeatureUnsupportedException instead of UnsupportedOperationException


herddb-jdbc/src/main/java/herddb/jdbc/HerdDBConnection.java
```diff
@@ -235,7 +235,7 @@ public void setCatalog(String catalog) throws SQLException {
 
     @Override
     public String getCatalog() throws SQLException {
-        return "default_catalog";
+        return null;
```
eolivelli, line 238: this was an error

herddb-jdbc/src/main/java/herddb/jdbc/HerdDBDatabaseMetadata.java
```diff
@@ -151,12 +151,12 @@ public boolean storesUpperCaseIdentifiers() throws SQLException {
 
     @Override
     public boolean storesLowerCaseIdentifiers() throws SQLException {
-        return false;
+        return true;
```
eolivelli, line 154: we are doing this way actually, so let's tell the truth

herddb-jdbc/src/main/java/herddb/jdbc/HerdDBDatabaseMetadata.java
```diff
@@ -176,12 +176,13 @@ public boolean storesLowerCaseQuotedIdentifiers() throws SQLException {
...
     public boolean storesMixedCaseQuotedIdentifiers() throws SQLException {
-        return false;
+        // important for OpenJPA
+        return true;
     }
 
     @Override
     public String getIdentifierQuoteString() throws SQLException {
-        return "'";
+        return "`";
```
eolivelli, line 185: we are doing this way actually, so let's tell the truth

eolivelli:
This work is in the direction of enhancing compatibility with Apache OpenJPA

herddb-jdbc/src/main/java/herddb/jdbc/HerdDBDatabaseMetadata.java
```diff
@@ -962,10 +961,10 @@ public ResultSet getColumns(String catalog, String schemaPattern, String tableNa
         // using "select * " in order to be compatible with old versions
         String query = "SELECT * FROM SYSCOLUMNS WHERE 1=1 ";
         if (tableNamePattern != null && !tableNamePattern.isEmpty()) {
-            query = query + " AND table_name LIKE '" + SQLUtils.escape(tableNamePattern) + "'";
+            query = query + " AND lower(table_name) LIKE '" + SQLUtils.escape(tableNamePattern.toLowerCase()) + "'";
```
diegosalvi, line 964: +1

PR #663, opened 2020-07-20
Issue 661: Use tablespace replication factor (expectedreplicacount) as configuration for BK write quorum size

eolivelli:
List of changes:
- use expectedreplicacount as minimum ensemblesize/writequorumsize/ackquorumsize
- allow to force ensemblesize/writequorumsize/ackquorumsize per-server, in order to allow a backward compatible configuration, this configuration will be a minimum ensemble size, if expectedreplicacount is greater that this value it will be ignored

### DIFF
--- a/herddb-core/src/main/java/herddb/core/TableManager.java
+++ b/herddb-core/src/main/java/herddb/core/TableManager.java
@@ -262,6 +262,8 @@ public final class TableManager implements AbstractTableManager, Page.Owner {
 
     private final boolean keyToPageSortedAscending;
 
+    private volatile boolean closed;
+
     void prepareForRestore(LogSequenceNumber dumpLogSequenceNumber) {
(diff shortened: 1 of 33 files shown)


eolivelli:
@diegosalvi please note this additional commit
https://github.com/diennea/herddb/pull/663/commits/1812cae1518bb3b497fc7b3ecc7bb8699bcba2ed#diff-d5e942a2119e7b01970ecec596ee19f3R297

herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -292,6 +294,12 @@ public long getUnloadedPagesCount() {
 
         @Override
         public long getTablesize() {
+            if (closed)  {
```
diegosalvi, line 297: good found

PR #670, opened 2020-08-04
Handle directly server side execution in application threads

eolivelli:
Follow up of #668 
When using LocalVMChannel we can execute server side peer code in the same thread of the application.

Changes:
- add new client option client.network.connect.localvm to disable connection to localvm server
- force server-to-server communication to use real network (we want tests to test something real)
- add a short-circuit path in LocalVMChannel to call directly the server peer
- server responses still run in the client callback threads (if we want to change this behavior it is better to do it in another patch)

### DIFF
--- a/herddb-core/src/main/java/herddb/client/HDBClient.java
+++ b/herddb-core/src/main/java/herddb/client/HDBClient.java
@@ -156,7 +156,7 @@ public class HDBClient implements AutoCloseable {
     Channel createChannelTo(ServerHostData server, ChannelEventListener eventReceiver) throws IOException {
         final int timeoutms = configuration.getInt(ClientConfiguration.PROPERTY_NETWORK_TIMEOUT, ClientConfiguration.PROPERTY_NETWORK_TIMEOUT_DEFAULT);
         final int timeouts = (int) TimeUnit.MILLISECONDS.toSeconds(timeoutms);
-        final boolean connectLocalVmAllowed = configuration.getBoolean(ClientConfiguration.PROPERTY_CLIENT_CONNECT_LOCALVM_SERVER, false);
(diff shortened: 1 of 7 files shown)


herddb-core/src/main/java/herddb/client/HDBClient.java
```diff
@@ -154,10 +154,16 @@ void releaseConnection(HDBConnection connection) {
...
 
     Channel createChannelTo(ServerHostData server, ChannelEventListener eventReceiver) throws IOException {
-        int timeoutms = configuration.getInt(ClientConfiguration.PROPERTY_NETWORK_TIMEOUT, ClientConfiguration.PROPERTY_NETWORK_TIMEOUT_DEFAULT);
-        int timeouts = (int) TimeUnit.MILLISECONDS.toSeconds(timeoutms);
-        return NettyConnector.connect(server.getHost(), server.getPort(), server.isSsl(), timeoutms, timeouts, eventReceiver, thredpool,
-                networkGroup);
+        final int timeoutms = configuration.getInt(ClientConfiguration.PROPERTY_NETWORK_TIMEOUT, ClientConfiguration.PROPERTY_NETWORK_TIMEOUT_DEFAULT);
+        final int timeouts = (int) TimeUnit.MILLISECONDS.toSeconds(timeoutms);
+        final boolean connectLocalVmAllowed = configuration.getBoolean(ClientConfiguration.PROPERTY_CLIENT_CONNECT_LOCALVM_SERVER, ClientConfiguration.PROPERTY_CLIENT_CONNECT_LOCALVM_SERVER_DEFAULT);
+        if (!connectLocalVmAllowed) {
```
pv3nturi, line 160: inverted if?

eolivelli: no, it isn't

if you allow LocalVMConnection then we you  `NettyConnector.connect` that detects if LocalVM is available.
if you do not allow LocalVMConnection then we force a network connection `NettyConnector.connectUsingNetwork`

PR #674, opened 2020-08-05
JDBC Implement getBoolean/setBoolean for integer columns

### LINKED ISSUE #672: JDBC: Handle ResultSet#getBoolean with integer values (1=true,0=false)
**BUG REPORT**

getBoolean() on a integer column returns always "false" because we are using Boolean.parseBoolean.

These are the official JDBC specs:

```


 /**
     * Retrieves the value of the designated column in the current row
     * of this <code>ResultSet</code> object as
     * a <code>boolean</code> in the Java programming language.
     *
     * <P>If the designated column has a datatype of CHAR or VARCHAR
     * and contains a "0" or has a datatype of BIT, TINYINT, SMALLINT, INTEGER or BIGINT
     * and contains  a 0, a value of <code>false</code> is returned.  If the designated column has a datatype
     * of CHAR or VARCHAR
     * and contains a "1" or has a datatype of BIT, TINYINT, SMALLINT, INTEGER or BIGINT
     * and contains  a 1, a value of <code>true</code> is returned.
     *
     * @param columnIndex the first column is 1, the second is 2, ...
     * @return the column value; if the value is SQL <code>NULL</code>, the
     * value returned is <code>false</code>
     * @exception SQLException if the columnIndex is not valid;
     * if a database access error occurs or this method is
     *            called on a closed result set
     */
```

eolivelli:
Fixes #672

### DIFF
--- a/herddb-core/src/main/java/herddb/sql/SQLRecordPredicate.java
+++ b/herddb-core/src/main/java/herddb/sql/SQLRecordPredicate.java
@@ -148,9 +148,6 @@ public class SQLRecordPredicate extends Predicate implements TuplePredicate {
                     }
                     return ((Number) value).longValue();
                 case ColumnTypes.DOUBLE:
-                    if (value instanceof Boolean) {
-                        return ((Boolean) value) ? 1d : 0d;
-                    }
                     return ((Number) value).doubleValue();
                 default:
                     return value;
--- a/herddb-core/src/main/java/herddb/sql/SQLRecordPredicate.java
+++ b/herddb-core/src/main/java/herddb/sql/SQLRecordPredicate.java
@@ -137,11 +137,20 @@ public class SQLRecordPredicate extends Predicate implements TuplePredicate {
             switch (type) {
                 case ColumnTypes.INTEGER:
                 case ColumnTypes.NOTNULL_INTEGER:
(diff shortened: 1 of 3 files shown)


herddb-jdbc/src/main/java/herddb/jdbc/HerdDBResultSet.java
```diff
@@ -235,7 +235,11 @@ public boolean getBoolean(int columnLabel) throws SQLException {
         fillLastValue(columnLabel);
         if (lastValue != null) {
             wasNull = false;
-            return Boolean.parseBoolean(lastValue.toString());
+            if (lastValue instanceof Number) {
```
eolivelli, line 238: this is actually the fix of getBoolean

herddb-core/src/main/java/herddb/sql/SQLRecordPredicate.java
```diff
@@ -137,11 +137,20 @@ public static Object cast(Object value, int type) {
             switch (type) {
                 case ColumnTypes.INTEGER:
                 case ColumnTypes.NOTNULL_INTEGER:
+                    if (value instanceof Boolean) {
```
eolivelli, line 140: this is the fix for setBoolean

PR #675, opened 2020-08-06
Implement more CAST cases

eolivelli:
Implement CAST operation for all of the expected cases.

The original error was:
` class herddb.utils.RawString cannot be cast to class java.lang.Number (herddb.utils.RawString is in unnamed module of loader 'app'; java.lang.Number is in module java.base of loader 'bootstrap')`

### DIFF
--- a/herddb-core/src/main/java/herddb/sql/SQLRecordPredicate.java
+++ b/herddb-core/src/main/java/herddb/sql/SQLRecordPredicate.java
@@ -148,9 +148,6 @@ public class SQLRecordPredicate extends Predicate implements TuplePredicate {
                     }
                     return ((Number) value).longValue();
                 case ColumnTypes.DOUBLE:
-                    if (value instanceof Boolean) {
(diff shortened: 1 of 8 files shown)


herddb-core/src/main/java/herddb/sql/SQLRecordPredicate.java
```diff
@@ -135,20 +135,110 @@ public static Object cast(Object value, int type) {
...
+                case ColumnTypes.NOTNULL_STRING:
+                    if (value instanceof RawString
+                            || value instanceof String) {
+                        return value;
+                    }
+                    if (value instanceof Number
+                            || value instanceof Boolean) {
+                        return value.toString();
+                    }
+                    throw new IllegalArgumentException("Value " + value + " (" + value.getClass() + ") cannot be converted to a Boolean");
```
dcdiennea, line 176: ...cannot be converted to a **String**

dcdiennea:
+1

PR #676, opened 2020-08-09
Diskless cluster mode v2

eolivelli:
This is a second version of the experimental diskless-cluster mode.

Recap about "diskless cluster mode":
With mode=diskless-cluster we are storing data pages and index pages on BookKeeper instead of the local disk.
In this mode the local disk is used only for tmpDirectory, to swap temporary resulsets (that do not survive to the restart of the server).
In order to store data on BK we created BookKeeperDataStorageManager, and generally this works this way:
- it mimics FiledataStorageManager behaviour
- metadata are stored on ZooKeeper
- real data (and index) pages are saved in BK Ledgers (one page per ledger....this is for the initial implementation, we could improve it in the future, but actually it makes it easy to perform deletion of data)
- the mapping from tablaspace,iddatapage -> idledger is stored on ZooKeeper
- in v1 we stored all of the metadata in znodes children of a znode with name equals to the nodeId of the server, this way each node was independent from the others
- in v1 leader and followers behave like in traditional "cluster" mode, in which followers tail the log and replay "locally" the data, this is not really "locally" because with BookKeeperDataStorageManager we are storing data to Bookies
- for instance in v1 when you have 3 nodes, each node handles datapages independently and this leads to a huge waste of space due the high level of write amplification (please also note that probably you will write to BK with 3/3/3 replication configuration, that is keeping 3 copies of each entry !)

Differences from v1:
- only the leader maintains the data, followers do nothing
- use ZooKeeper "multi" to ensure leadership for every write to data store metadata (mutl is like a 'transaction' in ZK, every operation must succeed in order to complete the whole operation, otherwise the batch does not have visible effects) 
- followers only start the TableSpaceManager and announce themselves on tablespacereplicastate
- allow 'replica:*' in CREATE/ALTER TABLESPACE command in order to allow ANY node to become replica and eventually leader in case of dead leader (you have to set maxleaderinactivitytime explicitly)
- add more test cases

Problems with v1 addressed by this PR:
- if you had 3 replicas with replicationfactor=3 you were going to have 9 copies of datapage !!
- there was no strong guarantee that only one node can write to the datapages
- there were no tests about the running of two "leaders" (the old leader MUST not be able to perform checkpoints or 'store' data to the storage layer) 
- there was no way to say "every node can handle this tablespace"

### DIFF
--- a/herddb-core/src/main/java/herddb/cluster/BookKeeperDataStorageManager.java
+++ b/herddb-core/src/main/java/herddb/cluster/BookKeeperDataStorageManager.java
@@ -1271,9 +1271,6 @@ public class BookKeeperDataStorageManager extends DataStorageManager {
     @Override
     public List<Table> loadTables(LogSequenceNumber sequenceNumber, String tableSpace) throws DataStorageManagerException {
         try {
-//            TODO
-//            String tableSpaceDirectory = getTableSpaceZNode(tableSpace);
-//            ensureZNodeDirectory(tableSpace, tableSpaceDirectory);
             String file = getTablespaceTablesMetadataFile(tableSpace, sequenceNumber);
             LOGGER.log(Level.INFO, "loadTables for tableSpace " + tableSpace + " from " + file + ", sequenceNumber:" + sequenceNumber);
             byte[] content = readZNode(file, new Stat());
@@ -1324,9 +1321,6 @@ public class BookKeeperDataStorageManager extends DataStorageManager {
     @Override
     public List<Index> loadIndexes(LogSequenceNumber sequenceNumber, String tableSpace) throws DataStorageManagerException {
         try {
-//          TODO
-//          String tableSpaceDirectory = getTableSpaceZNode(tableSpace);
-//          ensureZNodeDirectory(tableSpaceDirectory);
             String file = getTablespaceIndexesMetadataFile(tableSpace, sequenceNumber);
 
             LOGGER.log(Level.INFO, "loadIndexes for tableSpace " + tableSpace + " from " + file + ", sequenceNumber:" + sequenceNumber);
--- a/herddb-core/src/main/java/herddb/cluster/BookKeeperDataStorageManager.java
+++ b/herddb-core/src/main/java/herddb/cluster/BookKeeperDataStorageManager.java
@@ -643,6 +643,21 @@ public class BookKeeperDataStorageManager extends DataStorageManager {
         }
     }
 
+    private void deleteZNodeEnforceOwnership(String tableSpace, String path) throws DataStorageManagerException {
+        try {
+            zkWrites.inc();
+            Op opUpdateRoot = createUpdateTableSpaceRootOp(tableSpace);
+            Op opDelete = Op.delete(path, -1);
+            List<OpResult> multi = zk.ensureZooKeeper().multi(Arrays.asList(opUpdateRoot, opDelete));
+            updateRootZkNodeVersion(tableSpace, multi.get(0));
+        } catch (IOException | KeeperException err) {
+            throw new DataStorageManagerException(err);
+        } catch (InterruptedException err) {
+            Thread.currentThread().interrupt();
+            throw new DataStorageManagerException(err);
+        }
+    }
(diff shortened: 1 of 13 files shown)


eolivelli:
this is only a draft, because I did not still pushed points 3 (follower do nothing) and 4 (more tests)

eolivelli:
@dcdiennea @dmercuriali @amitvc @diegosalvi  PTAL

eolivelli:
@nicoloboschi I added more details as you requested offline, PTAL

herddb-core/src/main/java/herddb/cluster/BookKeeperDataStorageManager.java
```diff
@@ -1207,8 +1256,9 @@ private static LogSequenceNumber readLogSequenceNumberFromIndexMetadataFile(Stri
     @Override
     public List<Table> loadTables(LogSequenceNumber sequenceNumber, String tableSpace) throws DataStorageManagerException {
         try {
-            String tableSpaceDirectory = getTableSpaceZNode(tableSpace);
-            ensureZNodeDirectory(tableSpaceDirectory);
+//            TODO
```
diegosalvi, line 1259: Impl leftovers?

herddb-core/src/main/java/herddb/cluster/BookKeeperDataStorageManager.java
```diff
@@ -1259,8 +1309,9 @@ private static LogSequenceNumber readLogSequenceNumberFromIndexMetadataFile(Stri
     @Override
     public List<Index> loadIndexes(LogSequenceNumber sequenceNumber, String tableSpace) throws DataStorageManagerException {
         try {
-            String tableSpaceDirectory = getTableSpaceZNode(tableSpace);
-            ensureZNodeDirectory(tableSpaceDirectory);
+//          TODO
```
diegosalvi, line 1312: Impl leftovers?

herddb-core/src/main/java/herddb/model/TableSpace.java
```diff
@@ -81,6 +81,10 @@ private TableSpace(
         this.metadataStorageCreationTime = metadataStorageCreationTime;
     }
 
+    public boolean isNodeAssignedToTableSpace(String nodeId) {
+        return this.replicas.contains(nodeId) || this.replicas.contains("*");
```
diegosalvi, line 85: Could we use a constant?

diegosalvi:
I'm thinking that in the future a ledger cache on disk could be helpful

herddb-core/src/main/java/herddb/cluster/BookKeeperDataStorageManager.java
eolivelli: actually this is no more needed.
I am dropping this commented code, thanks

herddb-core/src/main/java/herddb/model/TableSpace.java
eolivelli: done

herddb-core/src/main/java/herddb/cluster/BookKeeperDataStorageManager.java
eolivelli: actually this is no more needed.
I am dropping this commented code, thanks

eolivelli:
> I'm thinking that in the future a ledger cache on disk could be helpful
@diegosalvi  can you elaborate more ?
This can be a good idea
can you file an issue with your idea please ?
this way we can address this comment in a follow up work

In my opinion copying data locally can help reads but it will add additional write amplification. So we should look for a good use case

eolivelli:
@diegosalvi this is the link to travis
https://travis-ci.org/github/diennea/herddb/builds/718589141

eolivelli:
latest link
https://travis-ci.org/github/diennea/herddb/builds/718591247

eolivelli:
Travis passed
https://travis-ci.org/github/diennea/herddb/builds/718591247

@diegosalvi your comments have been addressed. I am merging this patch now

PR #686, opened 2020-09-15
Introduce REPEATABLE_READ isolation level and default to READ_COMMITTED

eolivelli:
Since the beginning we always held 'read' lock in Transaction, in order to prevent modifications to a record 'SELECTED' within the context of a transaction.
This is far too restrictive than standard READ_COMMITTED isolation level.

This patch introduces the support for REPEATABLE_READ transaction isolation.
With REPEATABLE_READ the behavior is the same same as in 0.19.0 with READ_COMMITTED
With READ_COMMITTED (the default) each transaction won't keep the 'read' lock for each record

Results:
- "SELECT * FROM table" won't hold a lock for each record, saving lots of resources
- we will be nearer the expected behavior for SQL/JDBC

Motivation:
MagNews, that works with H2,SQLServer,Oracle and PostGRE assumes standard behavior and it is not able to work properly with the current behavior of HerdDB.

The patch also introduces an optimization regarding auto creation of transactions: if you do not have to keep read locks than it is useless to begin a transaction with a SELECT .... query, so we are going to postpone the creation of the transaction only at the first statemement that needs locks.

Notice:
This patch introduces a modification in the wire protocol, that in theory should be 100% compatible, but there is a bug in the server side processing of "Pdu#flags" field that actually makes it impossible to implement such 100% compatible change.
We will have to decide what to do before accepting this patch

### DIFF
--- a/herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
+++ b/herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
@@ -269,7 +269,7 @@ public class CalcitePlanner implements AbstractSQLPlanner {
         if (allowCache) {
             ExecutionPlan cached = cache.get(cacheKey);
             if (cached != null) {
-                return new TranslatedQuery(cached, new SQLStatementEvaluationContext(query, parameters, forceAcquireWriteLock));
+                return new TranslatedQuery(cached, new SQLStatementEvaluationContext(query, parameters, forceAcquireWriteLock, false));
             }
         }
 
@@ -323,7 +323,7 @@ public class CalcitePlanner implements AbstractSQLPlanner {
                         new SQLPlannedOperationStatement(values),
                         values
                 );
-                return new TranslatedQuery(executionPlan, new SQLStatementEvaluationContext(query, parameters, forceAcquireWriteLock));
+                return new TranslatedQuery(executionPlan, new SQLStatementEvaluationContext(query, parameters, false, false));
             }
 
             if (query.startsWith("SHOW")) {
@@ -357,7 +357,7 @@ public class CalcitePlanner implements AbstractSQLPlanner {
                     if (allowCache) {
                         cache.put(cacheKey, executionPlan);
                     }
-                    return new TranslatedQuery(executionPlan, new SQLStatementEvaluationContext(query, parameters, forceAcquireWriteLock));
+                    return new TranslatedQuery(executionPlan, new SQLStatementEvaluationContext(query, parameters, forceAcquireWriteLock, false));
                 }
             }
             if (maxRows > 0) {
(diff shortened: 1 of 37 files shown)


eolivelli:
TODO list:
- add tests,tests,tests,tests,tests,tests,tests
- handle the backward compatibility for the wire protocol
- implement GET

eolivelli:
added tests and implemented SELECT ... FOR UPDATE correctly

TODO:
- fix wire protocol
- add tests on JDBC

eolivelli:
created #687 for a fix on 0.19.x

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
```diff
@@ -323,7 +323,7 @@ public TranslatedQuery translate(String defaultTableSpace, String query, List<Ob
                         new SQLPlannedOperationStatement(values),
                         values
                 );
-                return new TranslatedQuery(executionPlan, new SQLStatementEvaluationContext(query, parameters, forceAcquireWriteLock));
+                return new TranslatedQuery(executionPlan, new SQLStatementEvaluationContext(query, parameters, false, false));
```
diegosalvi, line 326: Why this was forceAcquireWriteLock and now is always false?

herddb-utils/src/main/java/herddb/proto/Pdu.java
```diff
@@ -90,11 +92,11 @@ private Pdu(Recycler.Handle<Pdu> handle) {
     public long messageId;
 
     public boolean isRequest() {
-        return (flags | FLAGS_ISREQUEST) == FLAGS_ISREQUEST;
+        return (flags & FLAGS_ISREQUEST) == FLAGS_ISREQUEST;
```
diegosalvi, line 95: This was the PDU bug?

herddb-utils/src/main/java/herddb/proto/Pdu.java
```diff
@@ -90,11 +92,11 @@ private Pdu(Recycler.Handle<Pdu> handle) {
...
     }
 
     public boolean isResponse() {
-        return (flags | FLAGS_ISRESPONSE) == FLAGS_ISRESPONSE;
+        return (flags & FLAGS_ISRESPONSE) == FLAGS_ISRESPONSE;
```
diegosalvi, line 99: This was the PDU bug?

eolivelli: yes @diegosalvi

herddb-utils/src/main/java/herddb/proto/Pdu.java
eolivelli: yes @diegosalvi

herddb-core/src/main/java/herddb/sql/CalcitePlanner.java
eolivelli: `false` is the initial value.
It is not the planner that decides this flag but we are setting it is ServerSideConnectionPeer

diegosalvi (approved):

PR #689, opened 2020-09-17
Do not start callback threads for client in local mode

eolivelli:
The client callback threads are pretty useless in "local" mode, let's drop them by default


diegosalvi (approved):

eolivelli:
thank you @diegosalvi

PR #693, opened 2020-09-18
basic docker image setup (not created by default but ready to be created)

### LINKED ISSUE #470: Create official docker images for HerdDB
We should provide an official image for running HerdDB standalone and a cluster.

In standalone mode data is to be persisted on one or more volumes (data, journal, metadata)
In cluster mode we should coordinate with an existing ZooKeeper cluster, we can use the Embedded Bookie.
It is worth to think about not using the EmbeddedBookie as some @apache/pulsar users already have a BookKeeper cluster (and a ZooKeeper one)

rmannibucau:
Fixes #470


herddb-services/src/main/java/herddb/server/ServerMain.java
```diff
@@ -164,6 +172,15 @@ public void run() {
         }
     }
 
+    private static void useEnv() {
+        System.getProperties().putAll(System.getenv()); // legacy, to drop?
```
eolivelli, line 176: yes we can drop this line,
it was for the docker demo project

eolivelli (approved):
LGTM

PR #696, opened 2020-09-21
K8s skeleton yamls

rmannibucau:
Basic skeletons for K8s.

Idea will be to deploy bookkeeper and herddb in k8s in a custom managed fashion.

This PR is not complete and require the configuration of the link between both systems but gives some pointers about what is needed.

### DIFF
--- a/herddb-docker/src/main/kubernetes/samples/bookkeeper/4_bookkeeper.deployment.yaml
+++ b/herddb-docker/src/main/kubernetes/samples/bookkeeper/4_bookkeeper.deployment.yaml
@@ -1,0 +1,27 @@
+## from https://raw.githubusercontent.com/apache/bookkeeper/master/deploy/kubernetes/gke/bookkeeper.yaml
+
+##
+## Run BookKeeper auto-recovery from a different set of containers
+## Auto-Recovery makes sure to restore the replication factor when any bookie
+## crashes and it's not recovering on its own.
+##
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+  name: bookie-autorecovery
+spec:
+  replicas: 2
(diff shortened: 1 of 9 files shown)


herddb-docker/src/main/kubernetes/samples/bookkeeper/4_bookkeeper.deployment.yaml
```diff
@@ -0,0 +1,27 @@
...
+    metadata:
+      labels:
+        app: bookkeeper
+        component: bookkeeper-replication
+    spec:
+      containers:
+        - name: replication-worker
+          image: apache/bookkeeper:latest
+          command: [ "/bin/bash", "/opt/bookkeeper/entrypoint.sh" ]
+          args: ["/opt/bookkeeper/bin/bookkeeper", "autorecovery"]
```
eolivelli, line 24: this file is not needed for demo purposes,
this is a daemon that replicates data in case of lost bookies

if we want to make things simple you can drop this

rmannibucau: The templates are not for demo purposes but aims at giving a skeleton for real deployments IMHO.
For demo it is ok to use a plain ServerMain, no?

eolivelli: You are right, it should be better to have it.

We are not using autorecovery in production, as we usually have a small set of bookies and we are running autorecovery only manually and when needed (like when you lose definitively a machine)

so okay to keep this file now
thanks

rmannibucau:
no more relevant i guess

PR #698, opened 2020-09-22
JDBC: Do not depend on Netty Epool and Commons Pool2

eolivelli:
Changes:
- Use reflection and other classloading trick in order to not depend at runtime on Netty Epool and Commons Pool2
- Remove slf4j-jdk14 dependency (it should have been marked as 'test' scope)

JDBC Driver will work even with:
- netty epoll (it does not always bring performance improvements)
- commons pool2 (no need for pure JDBC Driver, connection pool is made by third party datasources)
- netty-codec (useless)
- netty tcboring-ssl (no need for "local" development or if you do not need TLS)

Motivation:
We want to shrink the JDBC driver Jar as much as possible


eolivelli:
Travis passed!
https://travis-ci.org/github/diennea/herddb/builds/729237961

nicoloboschi (approved):
+1

PR #699, opened 2020-09-22
Implement simple SQL commands with jSQLParser (experimental)

eolivelli:
This is a prototype about using jSQLParser for planning basic SQL commands:
- INSERT .... VALUES(x,x,x,x)
- SELECT col,col... FROM TABLE WHERE ... ORDER BY.....
- UPDATE TABLE set col=...... WHERE xxxxx
- DELETE FROM TABLE where .....

We are going to reuse all of the PlannerOp framework, this way:
- we are maintaining only one "implementation" of the data access system
- we can leverage all of the zero-copy optimizations that we introduced with Calcite

For "unsupported" operations we will fall back to Calcite.

Changes:
- add new configuration options: server.planner.type=calcite|jsqlparser|auto|none
- add systemproperty herddb.defaultplanner=calcilte (useful for CI)
- add Travis build with herddb.defaultplanner=jsqlparser
- Introduce fallback mechanism in case of server.planner.type=auto (try with jsqlparser, then fallback to Calcite)
- in case of server.planner.type=jsqlparser we want to be able to drop Calcite dependencies at runtime (@rmannibucau will be very happy!)
- implement INSERT/UPSERT/SELECT/UPDATE/DELETE statements
- implement GROUP BY (without Having)
- implement basic aggregations (COUNT,SUM....)
- implement GetStatement
- implement Indexes

Notable missing features:
- HAVING
- JOIN
- Handle automatic type inference (needed for some test case about string -> timestamp conversions)

Differences between the two Planners:

Calcite:
- slow, especially at boot time
- powerfull: it takes into account implicit Collactions (physical sort) of columsn, Nullable constranits, Table size, it simplifies expressions, it is able to elide aggregation and in generale to make the plan as efficient as possible
- handles mostly all SQL syntax
- bigger runtime (brings lots of third party dependencies)

jSQLParser:
- faster
- very short minded: very minimal similfications, not smart assumptions on dataset
- cannot handle all SQL syntax, especially JOINS, EXISTS...
- single jar runtime

### DIFF
--- a/herddb-core/src/main/java/herddb/core/DBManager.java
+++ b/herddb-core/src/main/java/herddb/core/DBManager.java
@@ -73,7 +73,9 @@ import herddb.server.ServerConfiguration;
 import herddb.server.ServerSidePreparedStatementCache;
 import herddb.sql.AbstractSQLPlanner;
 import herddb.sql.CalcitePlanner;
+import herddb.sql.DDLSQLPlanner;
 import herddb.sql.NullSQLPlanner;
+import herddb.sql.PlansCache;
 import herddb.storage.DataStorageManager;
 import herddb.storage.DataStorageManagerException;
 import herddb.utils.DefaultJVMHalt;
@@ -221,6 +223,9 @@ public class DBManager implements AutoCloseable, MetadataChangeListener {
         switch (plannerType) {
             case ServerConfiguration.PLANNER_TYPE_CALCITE:
                 planner = new CalcitePlanner(this, planCacheMem);
+                break;
+            case ServerConfiguration.PLANNER_TYPE_JSQLPARSER:
+                planner = new DDLSQLPlanner(this, new PlansCache(planCacheMem), new CalcitePlanner(this, planCacheMem));
                 break;
             case ServerConfiguration.PLANNER_TYPE_NONE:
                 planner = new NullSQLPlanner();
--- a/herddb-core/src/main/java/herddb/core/DBManager.java
+++ b/herddb-core/src/main/java/herddb/core/DBManager.java
@@ -220,12 +220,19 @@ public class DBManager implements AutoCloseable, MetadataChangeListener {
         preparedStatementsCache = new ServerSidePreparedStatementCache(statementsMem);
         String plannerType = serverConfiguration.getString(ServerConfiguration.PROPERTY_PLANNER_TYPE,
                 ServerConfiguration.PROPERTY_PLANNER_TYPE_DEFAULT);
+        PlansCache plansCache = new PlansCache(planCacheMem);
         switch (plannerType) {
             case ServerConfiguration.PLANNER_TYPE_CALCITE:
-                planner = new CalcitePlanner(this, planCacheMem);
+                planner = new CalcitePlanner(this, plansCache);
                 break;
             case ServerConfiguration.PLANNER_TYPE_JSQLPARSER:
-                planner = new DDLSQLPlanner(this, new PlansCache(planCacheMem), new CalcitePlanner(this, planCacheMem));
+                 // no Calcite fallback, you can drop Calcite dependecy
+                planner = new DDLSQLPlanner(this, plansCache, null);
+                break;
+            case ServerConfiguration.PLANNER_TYPE_AUTO:
+                // try with jSQLParser, fallback to Calcite
+                planner = new DDLSQLPlanner(this, plansCache,
+                        new CalcitePlanner(this, plansCache));
                 break;
             case ServerConfiguration.PLANNER_TYPE_NONE:
                 planner = new NullSQLPlanner();
--- a/herddb-core/src/main/java/herddb/sql/DDLSQLPlanner.java
+++ b/herddb-core/src/main/java/herddb/sql/DDLSQLPlanner.java
@@ -20,6 +20,7 @@
 package herddb.sql;
 
 import static herddb.sql.CalcitePlanner.findIndexAccess;
+import static herddb.sql.expressions.SQLParserExpressionCompiler.isBooleanLiteral;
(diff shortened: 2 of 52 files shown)


herddb-core/src/main/java/herddb/sql/DDLSQLPlanner.java
```diff
@@ -1142,6 +1171,83 @@ private Statement buildTruncateStatement(String defaultTableSpace, Truncate trun
         return new TruncateTableStatement(tableSpace, tableName);
     }
 
+    private ExecutionPlan buildInsertStatement(String defaultTableSpace, Insert insert, boolean returnValues) throws StatementExecutionException {
```
eolivelli, line 1174: This is the interesting part of this patch

herddb-core/src/main/java/herddb/sql/DDLSQLPlanner.java
```diff
@@ -1142,6 +1171,83 @@ private Statement buildTruncateStatement(String defaultTableSpace, Truncate trun
...
+        }
+        ItemsList itemsList = insert.getItemsList();
+        if (!(itemsList instanceof ExpressionList)) {
+            throw new StatementNotSupportedException();
+        }
+        List<Expression> projects = ((ExpressionList) itemsList).getExpressions();
+        int index = 0;
+        for (net.sf.jsqlparser.schema.Column column : insert.getColumns()) {
+            CompiledSQLExpression exp =
+                    SQLParserExpressionCompiler.compileExpression(projects.get(index));
```
eolivelli, line 1209: TODO: add "cast" for JDBCParameters, like Calcite does

herddb-core/src/main/java/herddb/sql/expressions/SQLParserExpressionCompiler.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+        if (expression instanceof JdbcParameter) {
+            JdbcParameter p = (JdbcParameter) expression;
+            return new JdbcParameterExpression(p.getIndex() - 1);
+        } else if (expression instanceof StringValue
+                || expression instanceof LongValue
+                || expression instanceof TimestampValue) {
+            return new ConstantExpression(DDLSQLPlanner.resolveValue(expression, false));
+        }
+//        else if (expression instanceof RexLiteral) {
```
eolivelli, line 50: TODO: implement all Expression types, I am leaving Calcite code, as far as test fail I am going to implement all of the supported expressions and operators

eolivelli:
@diegosalvi  if tests are passing then I feel this patch is good to go.
At least is is bringing in most of the change, then it is a metter to finish the implementation of basic missing features (HAVING, JOIN....)

herddb-core/src/main/java/herddb/core/DBManager.java
```diff
@@ -218,9 +220,19 @@ public Thread newThread(final Runnable r) {
         preparedStatementsCache = new ServerSidePreparedStatementCache(statementsMem);
         String plannerType = serverConfiguration.getString(ServerConfiguration.PROPERTY_PLANNER_TYPE,
                 ServerConfiguration.PROPERTY_PLANNER_TYPE_DEFAULT);
+        PlansCache plansCache = new PlansCache(planCacheMem);
         switch (plannerType) {
             case ServerConfiguration.PLANNER_TYPE_CALCITE:
-                planner = new CalcitePlanner(this, planCacheMem);
+                planner = new CalcitePlanner(this, plansCache);
+                break;
+            case ServerConfiguration.PLANNER_TYPE_JSQLPARSER:
```
amitvc, line 228: If a user creates some tables with calcite configuration and then switches back to using plain jsql parser will there be any issues? Would there be any issues?

eolivelli: No problem.
It is not about data, but only about how we access it.
Most probably it will use different plans.
For simple queries there won't be any difference, for complex queries probably Calcite will create better plans, but the results MUST be the same.

eolivelli:
@amitvc if you are happy with this patch can you please "approve" and merge it ?

thank you for your review

eolivelli:
I have pushed another commit:
- allow to drop Calcite-core and Avatica (still part of calcite)
- fix MySQL Backtick handling

herddb-core/src/main/java/herddb/sql/DDLSQLPlanner.java
```diff
@@ -230,31 +300,55 @@ public TranslatedQuery translate(
                 + ",query:" + query
                 + ",returnValues:" + returnValues
                 + ",maxRows:" + maxRows;
-        if (allowCache) {
-            ExecutionPlan cached = cache.get(cacheKey);
-            if (cached != null) {
-                return new TranslatedQuery(cached, new SQLStatementEvaluationContext(query, parameters, false, false));
+        try {
+            boolean forceAcquireWriteLock;
+            if (query.endsWith(" FOR UPDATE") // this looks very hacky
```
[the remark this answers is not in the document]
eolivelli: This is the same code in Calcite planner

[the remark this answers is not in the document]
eolivelli: there is not need to be picky here, if you don't write exactly "FOR UPDATE" at the end of the statement you will get an SQL error, so not a big deal.
This feature probably is not used that much, let's not waste CPU cycles :-)

diegosalvi (approved):

diegosalvi:
Good work!

eolivelli:
@amitvc We have assumed that you don't have any other comment and so we merged the patch.
Feel free to continue your review, I will be happy to fix any of your findings in the follow up work (I am going to work on JOINs now)

PR #702, opened 2020-09-30
Implement correctly alias handling in sort with jSQLParser

eolivelli:
Fix a bug in handling aliased column names in ORDER BY with simple jsqlparser based planner.

I have also moved most of SQL planning related tests to herddb.sql package in order to ease finding for those tests


eolivelli:
With this fix all tests should pass with jsqlparser  (please remember that tests about non supported feature are automatically excluded)

diegosalvi (approved):

PR #703, opened 2020-09-30
Fix handling of Planner Cache in case of leadership change

eolivelli:
While working on jSQLParser planner I noticed that Calcite Planner could be wrong in thinking about local availability of tablespaces, because it is caching "RootSchema".

With this change we are adding an additional check.
Changes:
- move AbstractSQLPlanner from interface to an abstract class
- implement the check
- fix a test case, that reproduced the issue only with jSQLParser

### DIFF
--- a/herddb-core/src/test/java/herddb/server/ClientMultiServerTest.java
+++ b/herddb-core/src/test/java/herddb/server/ClientMultiServerTest.java
@@ -225,6 +225,11 @@ public class ClientMultiServerTest {
                     } catch (ClientSideMetadataProviderException ok) {
                          assertTrue(ok.getCause() instanceof LeaderChangedException);
                     }
+
+                    // ensure that we don't use internal plans cache
+                    server_1.getManager().getPlanner().clearCache();
+                    server_2.getManager().getPlanner().clearCache();
+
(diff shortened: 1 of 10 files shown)


herddb-core/src/test/java/herddb/server/ClientMultiServerTest.java
```diff
@@ -225,6 +225,11 @@ public void requestMetadataRefresh(Exception error) throws ClientSideMetadataPro
                     } catch (ClientSideMetadataProviderException ok) {
                          assertTrue(ok.getCause() instanceof LeaderChangedException);
                     }
+
+                    // ensure that we don't use internal plans cache
```
eolivelli, line 229: without this fix the test passed on Calcite, but not with jSQLParser

eolivelli:
This issue is not yet stable, not worth to be committed now

diegosalvi (approved):

herddb-core/src/test/java/herddb/server/ClientMultiServerTest.java
diegosalvi: :O

PR #705, opened 2020-10-02
Cache ZooKeeper handle and invalidate cache in case of server network  address change

eolivelli:
I noticed that on the client we are not caching the ZooKeeper client ! This is a huge problem.
Also, in case of network address change of a node, the client is not invalidating propertly the routes maps and it also invalidate the caches too eagerly, requiring a great amount of work.

Changes:
- cache ZK client
- invalidate properly the caches


eolivelli:
This patch is not correct.
We are not closing the expired ZK handle.
We must also add tests about session expiration

nicoloboschi (approved):
+1 LGTM

Issue #678, opened 2020-09-01
Upgrade Apache Calcite to 1.26


aluccaroni:
This is a reminder of the task to update to Calcite 1.26 (when available):

- Checkout https://calcite.apache.org/news/releases/ regarding Calcite 1.25 and 1.26
- See https://github.com/apache/calcite/pull/2107 regarding "Enumerable Merge Join"
- Double check that no deprecated rule is in use (https://github.com/diennea/herddb/pull/666)

CC: @eolivelli @amitvc

PR #707, opened 2020-10-06
Remove JSONPath and Geometry third-party dependencies

eolivelli:
Apache Calcite requires a few third party dependencies in order to implement Geometry and JSON functions.
We don't need them but those classes are loaded at runtime.
The only way to workaround this issue is to create dummy classes in order to make Java happy

We are saving 1MB of space in the distribution ZIP and also we are no more adding these third-party deps to jdbc clients

Relates to #678


eolivelli:
@rmannibucau FYI

rmannibucau:
Should probably go in a dedicated module since there it breaks classpaths using these dependencies if herddb-core is there.
So my proposal would be to:

1. create a herddb-mock (or whatever name you like)
2. play with shades to have a classified version and keep default artifact "pure"

eolivelli:
@rmannibucau  
My initial idea was exactly the one you suggested.

I will update this patch

eolivelli:
@rmannibucau I have added more tweaks (remove useless commons-dbcp2 dep, keep only minimal part of Guava)

herddb-core/pom.xml
```diff
@@ -38,6 +38,11 @@
             <artifactId>herddb-utils</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>herddb-mock</artifactId>
+            <version>${project.version}</version>
```
rmannibucau, line 44: shouldnt it be scope=provided/optional (and only shade would be tweaked to include it)? By default grabbing core artifact I don't want it right?

herddb-jdbc/pom.xml
```diff
@@ -238,6 +246,17 @@
                                 </transformer>
                             </transformers>
                             <filters>
+                                <filter>
+                                    <artifact>com.google.guava:guava</artifact>
+                                    <includes>
```
rmannibucau, line 251: can need an IT test to ensure it still runs, a very simple "use calcite" test but otherwise it can be fragile with upgrades IMHO

herddb-core/pom.xml
eolivelli: good idea

eolivelli: good idea

herddb-jdbc/pom.xml
eolivelli: this is probably what you are looking for
https://github.com/diennea/herddb/blob/master/herddb-thirdparty/openjpa-test/pom.xml#L52

herddb-core/pom.xml
eolivelli: I am not rue about what it is better.
Currently we are not supporting Geometry and JSON functions so basically we don't need those library.
The only concern is about Applications that happen to need those jars.
In this case I guess they already have those jars on the classpath and they will have to explicitly exclude "herddb-mock".

Regarding the JDBC driver it is good to have this herddb-mock inside the JAR, properly relocated.

I am trying to add "optional" here now, and I am looking to the results

herddb-jdbc/pom.xml
rmannibucau: Can be, generally speaking I try to test in the same module to avoid unrelated surprises ("if openjpa fails it means openjpa module has a bug, not the jdbc shade"). But ok for me

eolivelli: I don't know how to test this artifact within the same module, do you mean with the maven-invoker-plugin ?
it would be kind of hard. I have never done this kind of tests before.
Do you have any example ?

Otherwise as you said, we can stay with current codebase, leveraging that module as a sort of integration test

rmannibucau: For example or just a failsafe/surefire binding on integration-test. @Test can just do a list in target to match the .jar (ensure to ignore -sources/-javadoc jars which happen in releases) and run it with a custom classloader for example.
Can be done later if too hard for this PR.

diegosalvi (approved):

PR #710, opened 2020-10-06
Miscellaneous fixes

eolivelli:
This is a list of little bugfixes found while porting MagNews.com from H2Database to HerdDB.

Tests are missing, so this is only a draft
- Implement CURRENT_DATE (as CURRENT_TIMESTAMP)
- reduce log for useless stuff
- handle Calcite DATE type (as Timestamp)
- dropindex IF EXISTS should be case insensitive

### DIFF
--- a/herddb-core/src/main/java/herddb/core/TableManager.java
+++ b/herddb-core/src/main/java/herddb/core/TableManager.java
@@ -2971,7 +2971,7 @@ public final class TableManager implements AbstractTableManager, Page.Owner {
             if (flushedDirtyPages > 0 || flushedSmallPages > 0 || flushedNewPages > 0 || flushedRecords > 0) {
                 LOGGER.log(Level.INFO, "checkpoint {0}, logpos {1}, flushed: {2} dirty pages, {3} small pages, {4} new pages, {5} records",
                     new Object[]{table.name, sequenceNumber, flushedDirtyPages, flushedSmallPages, flushedNewPages, flushedRecords});
-        
+
             }
 
             if (LOGGER.isLoggable(Level.FINE)) {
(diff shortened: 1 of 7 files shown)


herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -3014,11 +3017,13 @@ private TableCheckpoint checkpoint(
             result = new TableCheckpoint(table.name, sequenceNumber, actions);
 
             end = System.currentTimeMillis();
-
-            LOGGER.log(Level.INFO, "checkpoint {0} finished, logpos {1}, {2} active pages, {3} dirty pages, "
-                            + "flushed {4} records, total time {5} ms",
-                    new Object[]{table.name, sequenceNumber, pageSet.getActivePagesCount(),
-                            pageSet.getDirtyPagesCount(), flushedRecords, Long.toString(end - start)});
+            long totalTime = end - start;
+            if (totalTime > 100) {
```
diegosalvi, line 3021: Why logging based only on checkpoint duration?
Avoid to print a fast checkpoint or an empty checkpoint?
Print only if didn't flush records?

eolivelli: this is in order to not print useless checkpoints, like in "local" mode.
also it is useful to see if a checkpoint blocked table operations for a significant period of time.

In my opinion it is still good to see when checkpoints happen, even if much time passed since we saw checkpoint related problems.

using the number of records as threshold may result in hiding problems (the checkpoint is not doing anything) and probably it would also generate spam in "local mode" if the table has more than one dirty records

Do you have any suggestions ?

diegosalvi: I agree that is useful to know when checkpoint happens and how much time it took. My question was exactly about it because such check won't print checkpoint data if faster than 100 ms, it smells a little.
If we want to avoid useless checkpoint print it should't be logged only if it was empty (and perhaps fast too). Still in such case could be helpful a debug line

eolivelli: agreed.
I have checked the condition
PTAL @diegosalvi

diegosalvi (approved):

PR #713, opened 2020-10-08
Implement Oracle-LIKE DUAL table

eolivelli:
Implement a system table named 'dual' like Oracle specs:

`DUAL is a table automatically created by Oracle Database along with the data dictionary. DUAL is in the schema of the user SYS but is accessible by the name DUAL to all users. It has one column, DUMMY, defined to be VARCHAR2(1), and contains one row with a value X. Selecting from the DUAL table is useful for computing a constant expression with the SELECT statement. Because DUAL has only one row, the constant is returned only once. `

see https://docs.oracle.com/cd/B19306_01/server.102/b14200/queries009.htm

### DIFF
--- a/herddb-core/src/main/java/herddb/core/system/SysdualTableManager.java
+++ b/herddb-core/src/main/java/herddb/core/system/SysdualTableManager.java
@@ -47,7 +47,7 @@ public class SysdualTableManager extends AbstractSystemTableManager {
             .primaryKey("dummy")
             .build();
     private final List<Record> result;
-        
+
     public SysdualTableManager(TableSpaceManager parent) {
         super(parent, TABLE);
         this.result = Collections.singletonList(RecordSerializer.makeRecord(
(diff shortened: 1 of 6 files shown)


nicoloboschi (approved):
LGTM

herddb-core/src/main/java/herddb/core/system/SysdualTableManager.java
```diff
@@ -0,0 +1,64 @@
...
+            .primaryKey("dummy")
+            .build();
+
+    public SysdualTableManager(TableSpaceManager parent) {
+        super(parent, TABLE);
+    }
+
+    @Override
+    protected Iterable<Record> buildVirtualRecordList(Transaction transaction) {
+        List<Record> result = new ArrayList<>(1);
```
eolivelli, line 56: Actually we can create this list only once, in the constructor.
Let me fix it

diegosalvi: It could be a Collections#singletonList too

amitvc (approved):
LGTM

eolivelli:
@amitvc thanks for your review

Issue #497, opened 2019-11-04
Calcite - CLI - case sensitivy


aluccaroni:
**BUG REPORT**

1. Please describe the issue you observed:

It seems that we have some issue with the CLI parser (Calcite?) with case sensitivity.
See also https://github.com/diennea/herddb/pull/495.

In particular if I create a TABLE with a mix of upper and lowercase chars, then I'm unable to use the command "SHOW CREATE TABLE".

` CREATE TABLE mytablespace.q1_additional_info(msg_id long not null,sid integer not null,name string not null,value string not null,PRIMARY KEY(msg_id,sid,name))`

```
SHOW CREATE TABLE mytablespace.teSt10
Executing query:SHOW CREATE TABLE mytablespace.teSt10
ERROR:java.sql.SQLException: herddb.model.TableDoesNotExistException: Table teSt10 does not exist.
        at herddb.sql.CalcitePlanner.calculateShowCreateTable(CalcitePlanner.java:379)
        at herddb.sql.CalcitePlanner.translate(CalcitePlanner.java:287)
        at herddb.server.ServerSideConnectionPeer.handleOpenScanner(ServerSideConnectionPeer.java:445)
        at herddb.server.ServerSideConnectionPeer.requestReceived(ServerSideConnectionPeer.java:181)
        at herddb.network.netty.NettyChannel.lambda$handlePduRequest$0(NettyChannel.java:96)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.base/java.lang.Thread.run(Thread.java:834)
```

A part from fixing the issue at hand we need to check case sensitivity for all the CLI commands (I expect other commands to have problems).

amitvc:
@aluccaroni @eolivelli I will look into this tonight.

amitvc:
@aluccaroni @eolivelli Many places in our code we assume the query would be in upper case. 
[...]

PR #714, opened 2020-10-11
Fix to allow passing upper/lower case table names to the SHOW CREATE …

amitvc:
Summary of change - 
This change is for issue https://github.com/diennea/herddb/issues/497. Currently we are not able to pass a mix of upper/lower case table names to the SHOW CREATE TABLE command. This PR fixes that issue.

### DIFF
--- a/herddb-core/src/main/java/herddb/sql/functions/ShowCreateTableCalculator.java
+++ b/herddb-core/src/main/java/herddb/sql/functions/ShowCreateTableCalculator.java
@@ -124,6 +124,7 @@ public class ShowCreateTableCalculator {
                 tableName = query.trim();
             }
 
+            tableName = tableName.toLowerCase();
             TableSpaceManager tableSpaceManager = manager.getTableSpaceManager(tableSpace);
 
             if (tableSpaceManager == null) {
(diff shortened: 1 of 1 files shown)


amitvc:
@eolivelli - I am trying to add UT for this change but its hard to do so. The DBManager::getTableSpaceManager is not part of the interface that DBManager implements hence I cannot even pass the UT a Mocked version DBManager. I am trying to test ShowCreateTableCalculator::calculateShowCreateTable. I already have UTs for ShowCreateTableCalculator::calculate which is called from calculateShowCreateTable.

amitvc:
@eolivelli - Also this same use case happens for tablespace as well.  When we create tablespace we are preserving the case for the tablespace. Hence the SHOW CREATE TABLE fails if you pass a tablespace which is not exact match (upper or lower). We should probably fix that as well?

eolivelli (approved):
LGTM

eolivelli:
> @eolivelli - Also this same use case happens for tablespace as well.  When we create tablespace we are preserving the case for the tablespace. Hence the SHOW CREATE TABLE fails if you pass a tablespace which is not exact match (upper or lower). We should probably fix that as well?

Sure.
It is better to let the user work in a case insensitive way

PR #723, opened 2020-10-22
Validate alter table add fk

eolivelli:
- Validate FK constraint while issuing an ALTER TABLE ADD ... FOREIGN KEY ...
- Take current transaction into consideration
- Please remember that ALTER TABLE implicitly commits the transaction


eolivelli:
This a draft, missing test cases

diegosalvi (approved):

PR #727, opened 2020-10-28
More OpenJPA compatibility for DDL with backticks

eolivelli:
- Add tests for foreign keys and unique constrains as created by OpenJPA (see https://github.com/apache/openjpa/pull/76)
- Handle backticks in tablespace name on DDL statements

### DIFF
--- a/herddb-core/src/test/java/herddb/sql/RawSQLTest.java
+++ b/herddb-core/src/test/java/herddb/sql/RawSQLTest.java
@@ -2744,12 +2744,12 @@ public class RawSQLTest {
         String nodeId = "localhost";
         try (DBManager manager = new DBManager("localhost", new MemoryMetadataStorageManager(), new MemoryDataStorageManager(), new MemoryCommitLogManager(), null, null)) {
             manager.start();
-            assertTrue(manager.waitForBootOfLocalTablespaces(10000));            
+            assertTrue(manager.waitForBootOfLocalTablespaces(10000));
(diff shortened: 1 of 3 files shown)


herddb-core/src/test/java/herddb/sql/RawSQLTest.java
```diff
@@ -2738,6 +2739,25 @@ public void createIndexOnTableTest() throws Exception {
         }
     }
 
+    @Test
+    public void openJPATest() throws Exception {
+        String nodeId = "localhost";
+        try (DBManager manager = new DBManager("localhost", new MemoryMetadataStorageManager(), new MemoryDataStorageManager(), new MemoryCommitLogManager(), null, null)) {
+            manager.start();
+            assertTrue(manager.waitForBootOfLocalTablespaces(10000));
+            execute(manager, "CREATE TABLE `herd`.`parentTable` (`k1` VARCHAR NOT NULL, `p1` INTEGER)", Collections.emptyList());
```
diegosalvi, line 2748: Do we have a test on backtick for tablespace with DML?

eolivelli: yep, we have a few tests about backticks, but as we can see here we did not have them for "ALTER TABLE"

btw OpenJPA Dictionary will delimit every identifier, this is done because of the aggressive list of reserved words on Apache Calcite

diegosalvi (approved):

PR #728, opened 2020-10-28
Add OpenJPA foreignkey example

eolivelli:


eolivelli:
This is a draft as it depends on snapshot version of OpenJPA

herddb-thirdparty/openjpa-test/pom.xml
```diff
@@ -41,7 +41,7 @@
         <dependency>
             <groupId>org.apache.openjpa</groupId>
             <artifactId>openjpa</artifactId>
-            <version>3.1.1</version>
+            <version>3.1.3-SNAPSHOT</version>
```
eolivelli, line 44: this can be seen a problematic for a release, but it is needed in order to see the test work.
if we want to cut a release we must revert back to 3.1.2 and comment parts of the test
I hope that with @rmannibucau 's help we could release OpenJPA 3.1.3 soon

herddb-thirdparty/openjpa-test/src/main/resources/META-INF/persistence.xml
```diff
@@ -22,17 +22,29 @@
...
   </persistence-unit>
   
   <persistence-unit name="hdb_jdbc" transaction-type="RESOURCE_LOCAL">
     <class>test.entity.User</class>
+    <class>test.entity.Address</class>
     <exclude-unlisted-classes>true</exclude-unlisted-classes>
     <properties>
+      <property name="openjpa.Log" value="SQL=TRACE"/>      
+<!--      <property name="openjpa.jdbc.SchemaFactory" 
+          value="native(foreignKeys=true)" />
```
eolivelli, line 45: "native" creates real FKs but it does not work with "demiters", I will work on OpenJPA side in order to fix this problem

herddb-thirdparty/openjpa/src/main/java/herddb/openjpa/DBDictionary.java
```diff
@@ -28,9 +28,8 @@
     public DBDictionary() {
         platform = "HerdDB";
         databaseProductName = "HerdDB";
-        supportsForeignKeys = false;
-        supportsUniqueConstraints = false;
-        supportsCascadeDeleteAction = false;
+        supportsDeferredConstraints = false;
```
eolivelli, line 31: this is the same change did on OpenJPA 3.1.3-SNAPSHOT on "official" HerdDB dictionary

eolivelli:
For reference this is the bug on OpenJPA
https://issues.apache.org/jira/browse/OPENJPA-2837

I will work on it this week
cc @rmannibucau

eolivelli:
We have to wait for OpenJPA 3.1.3
I have filed this PR with the first part of the fix, without the OpenJPA 3.1.3 upgrade with support for FKs

PR #734, opened 2020-11-18
Move to slf4j and logback

eolivelli:
- Move from Java Util Logging to slf4j (with help from http://www.slf4j.org/migrator.html and NetBeans)
- Use LogBack for cli, tests and services package

### DIFF
--- a/herddb-cli/pom.xml
+++ b/herddb-cli/pom.xml
@@ -53,16 +53,16 @@
             <artifactId>jline</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-jdk14</artifactId>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>       
         </dependency>
-    </dependencies>
+        </dependencies>
     <build>
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
-                <version>3.2.1</version>
+                <version>3.2.4</version>
                 <executions>
                     <execution>
                         <phase>package</phase>
--- a/herddb-cli/src/main/java/herddb/cli/HerdDBCLI.java
+++ b/herddb-cli/src/main/java/herddb/cli/HerdDBCLI.java
@@ -24,6 +24,7 @@ import static herddb.client.ClientConfiguration.PROPERTY_ZOOKEEPER_ADDRESS_DEFAU
 import static herddb.client.ClientConfiguration.PROPERTY_ZOOKEEPER_PATH;
 import static herddb.client.ClientConfiguration.PROPERTY_ZOOKEEPER_PATH_DEFAULT;
 import static herddb.client.ClientConfiguration.PROPERTY_ZOOKEEPER_SESSIONTIMEOUT;
+import ch.qos.logback.classic.Level;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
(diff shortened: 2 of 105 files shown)


eolivelli:
@rmannibucau you have far more experience than us with logback and slf4j
can you please take a look to logback configuration ?

The rest of the patch is mostly a bulk update of the code, there are lots of places in which use placeholders but I wanted to make a minimal change

rmannibucau:
@eolivelli patch looks ok but have to admit using JUL was likely saner as a default. I would rather import the 2-3 classes of cxf doing the bridge (https://github.com/apache/cxf/blob/master/core/src/main/java/org/apache/cxf/common/logging/Slf4jLogger.java) than move to SLF4J first.

eolivelli:
@rmannibucau I am not sure I understand how to exploit https://github.com/apache/cxf/blob/master/core/src/main/java/org/apache/cxf/common/logging/Slf4jLogger.java

We have two main use cases:
- the server (standalone or embedded)
- the JDBC Driver/client

this patch is toward helping the users of the JDBC Driver that use slf4j, with this change the logs will flow into the expected log system.

if we add some kind of bridge from JUL to slf4j I don't know who can see benefits, the server users ?

eolivelli:
with this new way logback is used only by test cases and by the standalone server binary distribution

rmannibucau:
@eolivelli ok, I'm a JUL user (without SLF4J) so will you do another PR to drop slf4j? slf4j-jdk12 uses a new exception to fill the LogRecord data which is undesired on top of JUL. The opposite is true for slf4j users, you can add slf4j-jul bridge (or friends like log4j2 one). My (cxf link) proposal was to have a logger factory facade in herddb which returns JUL native Logger and subclassing logger as cxf did you can use slf4j,  log4j2 or any impl you want but stick to JUL as an API. Default impl would stay JUL but it would enable by a config to switch to slf4j if desired.

eolivelli:
@rmannibucau 
So basically you are saying that is it better to stick to JUL API in the code
and in every point that we are instantiating the Logger we can use  some utility method that uses pure JUL or creates a j.u.l.Logger implementation like https://github.com/apache/cxf/blob/master/core/src/main/java/org/apache/cxf/common/logging/Slf4jLogger.java

```
class Foo {
      private static final java.util.logging.Logger LOG = HerdDBLoggerFactory.makeLogger(Foo.class);
}

class HerdDBLoggerFactory {
       java.util.logging.Logger makeLogger(xxx) {
             if (usejul) {
               return ..; 
             }  else {
               return Slf4jWrapperLogger(...)
            }
     }
}
```

correct ?

rmannibucau:
@eolivelli tha's my opinion, you can get a last else which is opening the door to custom impl (as https://github.com/apache/openwebbeans/blob/master/webbeans-impl/src/main/java/org/apache/webbeans/logger/WebBeansLoggerFacade.java#L60). In terms of impl something along that can work well enough:

    if (classForced) return instantiate(classForced); // here you can also use ServiceLoader to test if there is some SPI (with priority) if you prefer.
    try { return  instantiate("name.of.Slf4jLoggerFactory"); }
    catch (Exception e) { return new JulLoggerFactory(); }

eolivelli:
okay I see your point @rmannibucau and I can adapt the patch to his proposal.

Personally I would prefer to move to slf4j:
- nicer API (especially the varargs)
- pluggable (but we could make it pluggable as @rmannibucau suggested)

IIUC the main problem about the slf4j-jul wrapper is about resource consumption and we are going to make less happy JUL users

@diegosalvi @nicoloboschi @mino181295 @amitvc what's your opinion ?

rmannibucau:
@eolivelli slf4j misses industrializable API log(level, record), a correct multiclassloader/env binding impl for allbindings (it is a singleton until the container handles it properly). Log4j2-api (which is a competitor) at least enables to solve both of issues. JUL is a good compromise cause always there and dep free (which is always good)

eolivelli:
@rmannibucau you convinced me :-) 
closing this PR now

PR #753, opened 2021-07-20
Implement first version of Change Data Capture support for HerdDB

eolivelli:
Add new utility to tail the HerdDB log from BookKeeper and create a flow
of the Mutations that happen inside a TableSpace.

```
try (ChangeDataCapture cdc = new ChangeDataCapture(
        tableSpaceUUID,
        clientConfiguration,
        new ChangeDataCapture.MutationListener() {
            @Override
            public void accept(ChangeDataCapture.Mutation mutation) {
                LOG.log(Level.INFO, "mutation "+mutation);
                mutations.add(mutation);
            }
        },
        LogSequenceNumber.START_OF_TIME);) {

      cdc.start();

      cdc.run();
 }
```

### DIFF
--- a/herddb-core/src/main/java/herddb/cdc/ChangeDataCapture.java
+++ b/herddb-core/src/main/java/herddb/cdc/ChangeDataCapture.java
@@ -286,7 +286,7 @@ public class ChangeDataCapture implements AutoCloseable {
             }
             break;
             case LogEntryType.COMMITTRANSACTION: {
-                TransactionHolder transaction = transactions.get(entry.transactionId);
+                TransactionHolder transaction = transactions.remove(entry.transactionId);
                 transaction.tablesDefinitions.forEach((tableName, tableDef) -> {
                     if (tableDef == null) { // DROP TABLE
                         tablesDefinitions.remove(tableName);
--- a/herddb-core/src/main/java/herddb/cdc/ChangeDataCapture.java
+++ b/herddb-core/src/main/java/herddb/cdc/ChangeDataCapture.java
@@ -19,11 +19,12 @@
  */
 package herddb.cdc;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
 
+import static herddb.client.ClientConfiguration.PROPERTY_ZOOKEEPER_ADDRESS;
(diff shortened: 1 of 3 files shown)


herddb-core/src/main/java/herddb/cdc/ChangeDataCapture.java
```diff
@@ -0,0 +1,320 @@
...
+                DataAccessorForFullRecord record = new DataAccessorForFullRecord(table, new Record(entry.key, entry.value));
+                fire(new Mutation(table, MutationType.UPDATE, record, lsn, entry.timestamp), entry.transactionId);
+            }
+            break;
+            case LogEntryType.BEGINTRANSACTION: {
+                transactions.put(entry.transactionId, new TransactionHolder());
+            }
+            break;
+            case LogEntryType.COMMITTRANSACTION: {
+                TransactionHolder transaction = transactions.get(entry.transactionId);
```
nicoloboschi, line 289: is it better to remove committed TransactionHolder object from the Map ? 
I feel this Map will grow forever

nicoloboschi (changes requested):
LGTM, I left a comment

herddb-core/src/main/java/herddb/cdc/ChangeDataCapture.java
eolivelli: good catch!
fixed

eolivelli:
I would like to add that currently this implementation works if you are able to follow the log from the beginning.
in case you are missing some part of the log we cannot make it work.
We cannot download a snapshot from the current leader because we need to replay all the mutations in the exact order and a snapshot is only a summary with the latest version of the "existing" records

eolivelli:
@diegosalvi @aluccaroni @nicoloboschi @amitvc any other comments ?

I have to follow up with another part about storing persistently all the versions of the Table schemas because when you restart the CDC Source you do not have the old versions of each Table.

But I prefer to follow up with another PR

amitvc:
LGTM @eolivelli . Can you also add steps on how to use this utility ?

eolivelli:
@amitvc currently this is only the first step, you can use the new API to connect to HerdDB and follow the changes on a tablespace.

My plan is to contribute a Pulsar IO Source Connector
https://pulsar.apache.org/docs/en/io-overview/

Issue #755, opened 2021-08-25
Improve memory configuration


diegosalvi:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

- permit to configure memory for data and keys as a percentage of max memory and not a fixed byte value
- permit to separate data and index memory like already done for pk memory

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

- should-have

3. Provide any additional detail on your proposed use case for this feature.

- simplify server setup and changes when jvm max heap is changed
- permit to separate data & indexes with large tables

diegosalvi:
Issue solved witg #757

PR #757, opened 2021-08-27
Improve memory configuration #755

diegosalvi:
This patch solves #755:

- add a new set of configuration to be able to configure memory constraints based on a percentage of maximum memory and not an absolute number: server.memory.data.percentage, server.memory.index.percentage server.memory.pk.percentage. Actually data and pk are simply a configurable value for a previous fixed one. The index is new, see below
- add a new set of configuration to enable indexes / data pages memory separation. Up untill now data and indexes was stored in the same memory space with the same constraint. Now you can configure server.memory.index.percentage (maximum memory percentage value, defaults to 0%) or server.memory.index.limit (byte fixed value, default to 0). If dedicated index memory resolves to a value more than 0 it will be used instead of data memory. By default index will continue to share data memory
- Due to memory inspections I noted that memory occupancy is evaluated enough well, we don't need to keep up to 50% memory free by default for "every else". Increased data default space from 30% to 50%.

 - [ x ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)


diegosalvi:
Some test failed, it seems only testing problems, checking them

eolivelli (approved):
LGTM

nicoloboschi (approved):
good work!

PR #766, opened 2021-12-24
Update to Calcite 1.28

eolivelli:
- Update Calcite to 1.28 (necessary because old versions of Calcite are flagged with CVEs)
- Implement (partial) handling of SEARCH (unhandled case should throw errors and not be executed with potentially wrong results)
- Implement handling of NlsString
- add @SuppressFBWarnings("NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE") as Calcite added some Nullable/NonNullable annotations that make most of the getTable() calls report an error in Spotbugs


nicoloboschi (approved):
Lgtm

dmercuriali (approved):

PR #774, opened 2022-01-03
Upgrade jetty version to 9.4.44.v20210927

hamadodene:


hamadodene:
At the moment we are unable to update to the latest version of jetty as we use Bookkeeper's PrometheusServlet.
However, we have that Bookkeeper uses jetty 9.43.x.
For now we update to 9.44

aluccaroni (approved):

diegosalvi (approved):

PR #777, opened 2022-01-10
Add back CLI to the docker image

eolivelli:
- Remove log4j2 references
- Unpack HerdDB Service zip in the docker image
- Use the /opt/herddb/bin/service script to start the server in the docker image

Most of the tutorial will work again, but not for "cluster" mode, as the way we are passing parameters changed, I will update the docs after merging this patch.

Basically you have to prepend `herddb_env_` to every entry

To start the service in standalone local mode:
Build the docker image:
```
mvn clean install -DskipTests
mvn package jib:dockerBuild@build -pl herddb-docker

```
Run:
`docker run --rm -it -p 7000:7000 -p 9845:9845 --name herddb herddb/herddb-server:0.24.0-SNAPSHOT  `
Connect to the CLI
`docker exec  -it herddb bash`
`bin/herddb-cli.sh`


herddb-services/src/main/resources/conf/log4j2.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Licensed to the Apache Software Foundation (ASF) under one
```
aluccaroni, line 3: We need to change the license header to the Diennea one

eolivelli:
@rmannibucau @aluccaroni @diegosalvi @dianacle @nicoloboschi @amitvc @dmercuriali this patch is still not perfect but I wanted to use herddb-services.zip in the docker image in order to have consistent packaging and in order to be sure that the docker image works exactly like the binary package.

I am not aware of any user nor of the .zip file neither of the docker image, so we can change how it works.
I would like to start promoting the docker image soon.
I hope that we will have soon some simple Helm Chart to deploy HerdDB on Kubernetes.
Also, Pulsar users already have a ZooKeeper and BooKeeper cluster on their k8s environment so starting HerdDB should be very simple

aluccaroni (changes requested):

rmannibucau:
Not sure log4j2 is a good choice, it is quite big and slow to start, what's the issue with JUL? worse case for docker yupiik-logging-jul is lighter and cloud friendly if you don't want to extend it.

eolivelli:
> Not sure log4j2 is a good choice

@rmannibucau you added log4j2 in the docker image, so I am simply moving forward with the work you did.

JUL works well for me, and it is the current logging provider used in herddb-services

rmannibucau:
@eolivelli hmm, have to admit I'm not sure/dont recall why I did that but think it is cause I copied a meecrowave setup more than anything else. What is required is a json output but between log4j2 support and jsonb+yupiik-logging-jul i prefer last one, in particular now.

eolivelli:
@rmannibucau are you okay if I revert and leave only JUL ?

rmannibucau:
+1, sorry about that

eolivelli:
@aluccaroni @rmannibucau I updated the patch PTAL

herddb-docker/pom.xml
```diff
@@ -149,8 +108,11 @@
               </to>
               <container>
                 <creationTime>USE_CURRENT_TIMESTAMP</creationTime>
-                <extraClasspath>${image.workdir}/custom/*</extraClasspath>
-                <mainClass>herddb.server.ServerMain</mainClass>
+                <entrypoint>/opt/herddb/bin/service</entrypoint>
```
rmannibucau, line 111: generally speaking it is always better to bypass scripts  since it ensures the process #1 is `java` and there is no issue in any container environment. With recent jib version, jib will ensure the classpath is in a file so it is set as `-cp @jib-classpath-file` so easy to override in k8s or any orchestrator when needed. Using a script always indirects it and can create issues if not well tested in the build (`exec`/`nohup` evolutions in particular) so is more risky IMHO.

rmannibucau (approved):
not critical since overridable but not sure the entry point should use cli/scripts instead of plain java command but looks good otherwise (in particular logger usage in the main ;)).

aluccaroni (approved):

PR #785, opened 2022-04-26
Trace data fetch page retry

diegosalvi:
This is just a log improvement. When receiving DataStorageManagerException for "moved" record we don't have any trace of attempted pages without setting the whole TableManager to FINE log.
Added a trace of attempted page and the next page to attempt (but aborted due to too many attempts) at the resulting exception message

### DIFF
--- a/herddb-core/src/main/java/herddb/core/TableManager.java
+++ b/herddb-core/src/main/java/herddb/core/TableManager.java
@@ -3838,7 +3838,7 @@ public final class TableManager implements AbstractTableManager, Page.Owner {
             Predicate predicate, StatementEvaluationContext context,
             Transaction transaction, LocalScanPageCache lastPageRead, boolean primaryIndexSeek,
             boolean forWrite, boolean acquireLock
-    ) {
+    ) throws DataStorageManagerException {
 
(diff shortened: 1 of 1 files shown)


herddb-core/src/main/java/herddb/core/TableManager.java
```diff
@@ -3965,13 +3966,29 @@ private Record fetchRecord(Bytes key, Long pageId, LocalScanPageCache localScanP
                 LOGGER.log(Level.FINE, "table " + table.name + ", activePages " + pageSet.getActivePages() + ", record " + key + " deleted during data access");
                 return null;
             }
+            if (trialPages == null) {
+                trialPages = new long[maxTrials];
+            }
+            trialPages[trialPages.length - maxTrials] = pageId;
             pageId = relocatedPageId;
             if (maxTrials-- == 0) {
```
dmercuriali, line 3974: shouldn't this be --maxTrials?

diegosalvi: Yes I missread the maxTrials-- check. It should be --maxTrials. To be conservative we'll change maxTrials from 2 to 3

PR #794, opened 2022-11-24
Upgrade libs

diegosalvi:

This is a routine pr to upgrade some updated libraries.
Upgrades jackson-databind with vulnerability issues https://github.com/diennea/herddb/security/dependabot?q=package%3Acom.fasterxml.jackson.core%3Ajackson-databind+manifest%3Apom.xml+has%3Apatch


eolivelli (approved):
Lgtm

Usually it would be better to make separate PRs.

In case of problems it is hard to find which is the dependency and git revert only the problematic one.

Ci passed, it is a good sign that we work broke anything

PR #797, opened 2023-01-16
Replace fat cli with a light one

diegosalvi:
This is a draft to create a slim cli while keeping "fat" cli artifact


eolivelli (changes requested):
Makes sense to me

It is quite hard for the user to understand why have two CLI versions.

I guess that the problem is about the fact that by default from the CLI you can start the embedded server. 
I don't think that this is something that is expected.
I would simply fix the original CLI.
if you need to start a server over existing data you can download the zip distribution

aluccaroni:
+1 to simplify the cli and keep only one, even if it's a breaking change

eolivelli (changes requested):
As @aluccaroni said, it is better to have only one CLI tool and not two.
It would be confusing to have two packages

diegosalvi:
Sounds good to me

diegosalvi:
After tests this could be merged

herddb-cli/pom.xml
```diff
@@ -59,31 +59,6 @@
     </dependencies>
     <build>
         <plugins>
-            <plugin>
```
eolivelli, line 62: I think that if you drop this plugin the .jar won't be able to work.

We still need a uber jar, but we can remove anything that is not required to run the CLI.
I think that most of the weight is due to the rocksdb jar.
I would drop useless jars, like we did with the "thin" JDBC driver

eolivelli (changes requested):

herddb-cli/pom.xml
diegosalvi: The shaded jar generate various problems with his shaded dependencies.
You can just download the services and use the cli with it or in a more complex project run it with your other libraries.
Shading every dependency has many classloading problems when you want to upgrade some dependency (and then for some "strange" behavior you find loaded old classes from the shaded lib)

eolivelli: The Uber jar is not intended to be used as dependency but only to run the CLI from the command line.

I agree that we can let users use the zip file.
Also, now that we have docker images, it is super easy to run docker run -it .....

eolivelli (commented):
Do we have test Cades that very that the CLI packaged in the zip distro works?

Did you try it manually?

diegosalvi:
Good catch! I forgot the cli shell and bat files in services.
Fixed services package:
* changed scripts to use java -cp instead of java -jar
* moved herddb-cli to lib instead of main directory (it's cleaner this way)

dmercuriali (approved):
Tested locally. LGTM

eolivelli (approved):
LGTM
nice work !

PR #799, opened 2023-01-16
Cleanup local channel memory, fixes #798

### LINKED ISSUE #798: Local channels are never removed from LocalVMChannelAcceptor memory
**BUG REPORT**

- What did you do?
Just used client with local network enabled

- What did you expect to see?
local channels should be removed from memory when closed

- What did you see instead?
LocalVMChannelAcceptor keep a LocalVMChannel memory to force close them on acceptor close *but* such memory isn't cleaned in any other means. While using local connection such objects keep to be added to channels sets but they are never removed. Looking into a memory dump in production server I got 5GB retained from such list
Schermata 2023-01-16 alle 12 42 39
Referenced channels are closed but still there

diegosalvi:
Fixes #798.

Added LocalVMChannel deregistration method from LocalVMChannelAcceptor to avoid to keep stacking old channels without be able to reclaim their memory.

Added a test check to verify channels deregistration
Modified AbstractChannel close procedure to use an atomic boolean to ensure only once doClose invocation

### DIFF
--- a/herddb-net/src/main/java/herddb/network/netty/AbstractChannel.java
+++ b/herddb-net/src/main/java/herddb/network/netty/AbstractChannel.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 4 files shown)


eolivelli (approved):
LGTM

nice catch !

dmercuriali (approved):

hamadodene (approved):

PR #803, opened 2023-03-07
[security] Implement basic JWT authentication

eolivelli:
Motivation:
Implement JWT validation, this way you don't need a user database, but the administrator can issue "tokens" that store a set of "claims".

Modifications:
Implement basic JWT validation, that works only with USERNAME/PASSWORD SASL authentication, the password is the JWT token.

### DIFF
--- a/herddb-core/pom.xml
+++ b/herddb-core/pom.xml
@@ -139,6 +139,18 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>io.jsonwebtoken</groupId>
+            <artifactId>jjwt-api</artifactId>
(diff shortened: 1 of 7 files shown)


eolivelli:
This is only POC at this stage, missing parts:
- tests, tests, tests
- CLI tools to issue tokens (like the "pulsar tokens" CLI tool)

eolivelli:
Instead of building another CLI for JWT I am going to try this:
https://github.com/mike-engel/jwt-cli

Then we can write instructions on the docs about how to create tokens using it, without reinventing the wheel and also not having to maintain such kind of security related tools

PR #805, opened 2023-05-25
Update bookkeeper to 4.16

dmercuriali:
Updated bk dependency to 4.16
Use the new EmbeddedServer feature for the embedded bk server

### DIFF
--- a/herddb-core/pom.xml
+++ b/herddb-core/pom.xml
@@ -126,6 +126,7 @@
         <dependency>
             <groupId>commons-cli</groupId>
             <artifactId>commons-cli</artifactId>
+	        <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.hadoop</groupId>
--- a/herddb-core/pom.xml
+++ b/herddb-core/pom.xml
@@ -126,7 +126,6 @@
         <dependency>
             <groupId>commons-cli</groupId>
             <artifactId>commons-cli</artifactId>
-            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.hadoop</groupId>
--- a/herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
+++ b/herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
@@ -41,6 +41,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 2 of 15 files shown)


herddb-core/src/main/java/herddb/cluster/EmbeddedBookie.java
```diff
@@ -188,16 +190,17 @@ private void dumpBookieConfiguration(Path bookie_dir, org.apache.bookkeeper.conf
...
+        if (embeddedServer != null) {
             LOG.info("Apache Bookkeeper stopping");
             try {
-                bookieServer.shutdown();
-
-                bookieServer.join();
+                embeddedServer.getLifecycleComponentStack().close();
+                if (waitForBookieServiceState(Lifecycle.State.CLOSED)) {
+                    LOG.info("Apache Bookkeeper stopped");
+                }
```
eolivelli, line 199: what about adding a log line in case of timeout ?

dmercuriali: added a warning here and in the start() method

eolivelli (approved):

herddb-core/pom.xml
```diff
@@ -126,7 +126,6 @@
         <dependency>
             <groupId>commons-cli</groupId>
             <artifactId>commons-cli</artifactId>
-            <scope>test</scope>
```
diegosalvi, line 129: Why import commons-cli not only in tests?

diegosalvi (approved):

herddb-core/src/main/java/herddb/cluster/BookkeeperCommitLog.java
```diff
@@ -246,8 +249,8 @@ public void close() throws LogNotAvailableException {
             try {
                 LOGGER.log(Level.INFO, "{0} closing ledger {1}, with LastAddConfirmed={2}, LastAddPushed={3} length={4}, errorOccurred:{5}",
                         new Object[]{tableSpaceDescription(), out.getId(), out.getLastAddConfirmed(), out.getLastAddPushed(), out.getLength(), errorOccurredDuringWrite});
-                out.close();
-            } catch (InterruptedException | org.apache.bookkeeper.client.api.BKException err) {
+                out.closeAsync().get(60, TimeUnit.SECONDS);
+            } catch (InterruptedException | ExecutionException | TimeoutException err) {
```
eolivelli, line 253: usually it i s better to unwrap ExecutionException.getCause() in order to get rid of the intermediate ExecutionException.

We should also make the timeout configurable

eolivelli (commented):
Nice work !

I have added some last minute feedback

@diegosalvi @dmercuriali

eolivelli:
Also there is an ongoing VOTE for BK 4.16.2, I suggest to wait for that release before cutting the HerdDB release

cc @aluccaroni

dmercuriali:
follow up PR #810

